PR #21, opened 2021-02-05
[WIP] Additional kernels

jejjohnson:
This pull request will try to merge all changes of the additional kernels branch to the main branch.

Features the following kernels:
* [ ] Linear Kernel
* [X] Matern1/2 Kernel
* [X] Matern3/2 Kernel
* [X] Matern5/2 Kernel
* [ ] RationalQuadratic Kernel
* [ ] Polynomial kernel

**Note**: the Matern52 fails for the PSD matrix test. Even with Jitter. The eigs are negative and there are complex values...


gpjax/kernels/base.py
```diff
@@ -0,0 +1,55 @@
...
+    Base class for all kernel functions. By inheriting the `Module` class from Objax, seamless interaction with model parameters is provided.
+    """
+    def __init__(self, name: Optional[str] = None):
+        """
+        Args:
+            name: Optional naming of the kernel.
+        """
+        self.name = name
+        self.spectral = False
+
```
thomaspinder, line 19: It's good that we've removed the `Stationary` inheritance, but it may be beneficial to have stationarity as a property of the kernel class.

Issue #12, opened 2021-01-02
Clean up typing of Jax arrays


thomaspinder:
There are currently a few different references to the same type i.e. `jnp.ndarray`, `jnp.array`, and `JaxArray`. This should be tidied up to ensure parity across all types. I'd suggest using `JaxArray` type as the default array type.

PR #22, opened 2021-02-05
Consistent Dtypes, Formatting and Dev Install

jejjohnson:
**Contrib I**:

Cleaned up package to ensure all `JaxArray` calls are consistent with `objax.types.JaxArray`. I did not change the typing in the notebook demonstrations. References #12. 

---
**Contrib II**:
Ran some formatting with `isort` and `black`. I created some Makefiles to help devs with having the necessary packages installed. Later this can be automated with the .github actions. Also, we can also put a short contrib guide using for future contributors.

---
**Contrib III**:
Added some key packages to help with devs in the `setup.py` file. This allows us not to keep extra packages in the requirements files for users who don't need it. One can simple do `pip install -e ".[dev]"` to install the development packages or `pip install -e ".[all]"` to install the requirements for everything. In the future, we can have an extra options like "notebooks".

---
**Contrib IV**:
I changed the relative-level imports to absolute level imports. It's my preference due to difficulties I have with relative imports in my dev environments. but I can revert the changes if request!


thomaspinder:
This is excellent. Thanks a lot for sorting all this out @jejjohnson!

PR #28, opened 2021-03-18
[WIP] Added preliminary numpyro integration.

jejjohnson:
A simple interface between GPJAX and numpyro. This will allow users to offload the inference to numpyro and they can focus on their models. It mainly acts as a starter pack because once models get more complex (e.g. deep GPs, latent variable models) then the user will need to dig in.

The demo notebook can be found [here](https://github.com/jejjohnson/GPJax-1/blob/numpyro/docs/nbs/numpyro_interface.ipynb). I will keep it updated to track progress.

---
#### Contributions

* Relatively simple interface between GPJax params and numpyro params - a simple nested dictionary.
* Simple way to add constraints to the nested dictionary - used multiple dispatch for a few cases (probably overkill)
* MLE estimation function using numpyro - uses `numpyro.param` 
* MAP estimation function using numpyro - uses `numpyro.distributions`
* Demo inference schemes - VI
* Demo sampling scheme - NUTS
* Showcase predictions and sampling from the posterior.

---
#### Todo

* [X] numpyro parameter conversion
* [X] Tests
* [X] numpyro prior conversion
* [X] Tests
* [X] regression
  * [X] numpyro mle/map (params/priors)
  * [x] Tests

---
#### Demo Notebook

The demo notebook will cover the following:

* [x] well commented...
* [X] how to convert to numpyro parameters
* [X] how to add numpyro constraints
* [X] how to declare numpyro samples
* [X] maximum likelihood estimation
* [X] maximum a priori estimation
* [x] Variational Inference
* [X] Monte Carlo (NUTS) sampling
* [x] predictions via the posterior samples

---
#### Observations

* It's somewhat difficult to cover all of the cases for ways to add `numpyro` constrains to the parameters. In the end, using `numpyro.distributions` is much cleaner (and probably recommended). I wouldn't try to put too much effort into this.
* While I focused on `numpyro` syntax, in principle, the `tfp` package is already integrated into the `numpyro` library. (See [here](http://num.pyro.ai/en/stable/distributions.html#tensorflow-distributions)). However, there is a noticeable difference in speed between the `tfp` distributions and `numpyro` distributions... `numpyro` is faster and I don't know why.
* The notebook is quite long and I'm only covering regression. Perhaps I should have a separate one for classification...?

---
#### Things for Later

* A way to have a mix of parameters and distributions...


gpjax/interfaces/numpyro.py
```diff
@@ -0,0 +1,129 @@
...
+def add_priors(params: dict, priors: numpyro_priors):
+    new_params = deepcopy(params)
+    for ivariable in new_params.keys():
+        new_params[ivariable] = {"param_type": "prior", "prior": priors}
+    return new_params
+
+
+@dispatch(ConjugatePosterior, dict)
+def numpyro_marginal_ll(gp: ConjugatePosterior, numpyro_params: dict) -> Callable:
+    def mll(x: Array, y: Array):
```
thomaspinder, line 99: Change args to `data: Dataset`.

thomaspinder: First line of MLL fn can then just be `x, y = data.X, data.y`, meaning the rest of the function can be left unchanged.

tests/interfaces/test_numpyro/test_numpyro_marginal_ll.py
```diff
@@ -0,0 +1,292 @@
...
+    x = jr.normal(key=KEY, shape=(n_samples, n_features))
+    y = jr.normal(key=KEY, shape=(n_samples, n_latents))
+    return x, y
+
+
+def test_numpyro_marginal_ll_params():
+
+    # create sample data
+    X, y = _gen_training_data(10, 10, 2)
+
```
thomaspinder, line 45: Line here to instantiate `Dataset` using `X` and `y.

tests/interfaces/test_numpyro/test_numpyro_marginal_ll.py
```diff
@@ -0,0 +1,292 @@
...
+    # convert to numpyro-style params
+    numpyro_params = numpyro_dict_params(params)
+
+    # initialize numpyro-style GP model
+    npy_model = numpyro_marginal_ll(posterior, numpyro_params)
+
+    # do one forward pass with context
+    with numpyro.handlers.seed(rng_seed=KEY):
+        model_params = numpyro.handlers.trace(npy_model).get_trace(X, y.T)
```
thomaspinder, line 57: This will then become `Dataset` in the `get_trace()` call.

gpjax/interfaces/numpyro.py
jejjohnson: Thanks! Fixed.

tests/interfaces/test_numpyro/test_numpyro_marginal_ll.py
jejjohnson: Fixed!

tests/interfaces/test_numpyro/test_numpyro_marginal_ll.py
jejjohnson: Fixed!

PR #36, opened 2021-08-27
Added missing prior mean at test inputs

bodin-e:
Hi! 

First of all, I really like your library! It has a nice and well decoupled code structure.

On to the pull request:
I noticed that the mean function (evaluated at the test inputs) is not added to the output of the predictive mean function of the conjugate posterior, which I reckon is a bug. 

Are you happy with the following fix?


thomaspinder:
Hey! 

Thanks for the kind words! 

Nice catch with this! You're correct - this is a bug so I'm happy to merge this PR.

Thanks again for your contribution!

PR #55, opened 2022-05-20
Revise fit in abstractions & minor documentation update.

Daniel-Dodd:
Revise abstraction's `fit` to only support optax optimisers. 

This update involved:

- Removing the at present function `fit` from abstractions
- Renaming `optax_fit` to `fit` to act as the new function that only supports optax
- Revising documentation notebooks to accommodate the new `fit` function & fix minor typos


Daniel-Dodd:
Need to update the unit tests.

PR #74, opened 2022-06-13
Add method to combine datasets.

Daniel-Dodd:
Use `__add__` method on `Dataset` object to facilitate combining two datasets (useful for optimal design purposes).

Example:
```python
import jax.numpy as jnp
import gpjax as gpx

x1 = jnp.array([[1., 2.], [3., 4.])
y1 = jnp.array([[2.], [5.]])

D1 = gpx.Dataset(X=x1, y=y1)

x2 = jnp.array([[7., 11.], [1., 3.], [5., 5.]])
y2 = jnp.array([[8.], [3.], [7.]])

D2 = gpx.Dataset(X=x2, y=y2)

D_combined = D1 + D2
```

- [ ] Bugfix
- [ X] Feature
- [ ] Build related changes

### DIFF
--- a/gpjax/types.py
+++ b/gpjax/types.py
@@ -24,7 +24,7 @@ class Dataset:
         )
 
     def __add__(self, other: "Dataset") -> "Dataset":
-        """Combines two datasets into one."""
+        """Combines two datasets into one. The right-hand dataset is stacked beneath left."""
         x = jnp.concatenate((self.X, other.X))  
         y = jnp.concatenate((self.y, other.y))  
 
--- a/gpjax/types.py
+++ b/gpjax/types.py
@@ -23,6 +23,13 @@ class Dataset:
             f"- Number of datapoints: {self.X.shape[0]}\n- Dimension: {self.X.shape[1]}"
         )
 
+    def __add__(self, other: "Dataset") -> "Dataset":
+        """Combines two datasets into one."""
+        x = jnp.concatenate((self.X, other.X))  
+        y = jnp.concatenate((self.y, other.y))  
+
+        return Dataset(X=x, y=y)
+
     @property
     def n(self) -> int:
         """The number of observations in the dataset."""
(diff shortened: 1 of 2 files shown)


tests/test_types.py
```diff
@@ -21,6 +22,20 @@ def test_dataset(n, outd, ind):
     assert d.in_dim == ind
     assert d.out_dim == outd
 
+    # test combine datasets
+    x2 = 2*jnp.ones((n2, ind))
+    y2 = 2*jnp.ones((n2, outd))
+    d2 = Dataset(X=x2, y=y2)
+
+    d_combined = d + d2
```
thomaspinder, line 30: Does this operation commute? i.e., `d+d2 == d2+d`?

gpjax/types.py
```diff
@@ -23,6 +23,13 @@ def __repr__(self) -> str:
             f"- Number of datapoints: {self.X.shape[0]}\n- Dimension: {self.X.shape[1]}"
         )
 
+    def __add__(self, other: "Dataset") -> "Dataset":
+        """Combines two datasets into one."""
```
thomaspinder, line 27: Can we just be explicit here that the right-hand dataset is stack beneath the left-hand dataset?

thomaspinder (changes requested):

thomaspinder:
Cool idea. Made two minor comments. Happy to merge once addressed.

Not an immediate priority, but it might be worth thinking about a notebook demonstrating this functionality. Perhaps a simple BO example.

tests/test_types.py
Daniel-Dodd: No. This is inconsistent due to stacking. This would require sorting the dataset into order.

gpjax/types.py
Daniel-Dodd: Good idea. Will expand the description.

tests/test_types.py
thomaspinder: Good point. We don't want to sort.

Daniel-Dodd:
> 

This would be nice. Happy to do this.

PR #87, opened 2022-07-20
Initial kronecker work

thomaspinder:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

### DIFF
--- a/gpjax/covariance_operator.py
+++ b/gpjax/covariance_operator.py
@@ -1,0 +1,250 @@
+# Copyright 2022 The GPJax Contributors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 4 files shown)


thomaspinder:
Comment from Dan's push:

- Minimal Kronecker kernel. The inverse is highly inefficient - this can be improved!
- Added white noise kernel.
- Added diagonal computation abstraction.
- Added minimal computation abstraction for kernel combinations (i.e., for * and +).

Main issues:
- Though we can replace the computations in the gps.py, we might miss out with caching stuff e.g., the Cholesky decomposition in the dense computation case. Is there a sensible way around this?
- As a further point to the preceding point, we should not be computing the gram matrix in general (yes I known we need to in the dense setting).

- Think about computation savings for KL divergences - how can we resolve these?

Minor issues:
- Need to discuss the best way to add some computation methods - it might be tidier than just adding a load of static methods altogether.
- Need tests.

Daniel-Dodd:
Closed PR - Kronecker kernels will be added from a different branch, and not `kernel_compute`.

PR #95, opened 2022-08-10
Fix conjugate regression bug for a single datapoint.

Daniel-Dodd:
Thanks @krunolp for raising this issue.

Conjugate regression contains a bug that arises when calculating the marginal log-likelihood on a single data point. In particular, fitting a GP using only single input-output pair [both with shape (1,1)] results in the following error:

```
bijector BlockShift has event_ndims_out==1, but the input has only 0 array dimensions. 
```

This PR fixes this bug and updates the tests for this edge-case. 

## Pull request type

- [X ] Bugfix
- [ ] Feature
- [ ] Build related changes

### DIFF
--- a/gpjax/gps.py
+++ b/gpjax/gps.py
@@ -234,7 +234,8 @@ class ConjugatePosterior(AbstractPosterior):
 
             constant = jnp.array(-1.0) if negative else jnp.array(1.0)
             return constant * (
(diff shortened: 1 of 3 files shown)


krunolp:
Thank you for looking into this!

thomaspinder (approved):

thomaspinder:
Looks good to me. Thanks for picking up on this!

Issue #96, opened 2022-08-11
bug: fit_batches training loop


Daniel-Dodd:
# Bug Report

**GPJax version: 0.4.10, 0.4.9**

**Current behavior:**

There is an issue regarding `fit_batches`. I am still diagnosing this, but I believe the TensorFlow dataset iterator (for mini-bathing) has issues within the `lax.scan`. I noticed this because I am unable to reproduce previous experimental results (from Easter) with older GPJax code. Furthermore, plotting the ELBO results in a smooth curve, which is not expected during stochastic optimisation, and the final model predictions are poor. Each training step appears to cache and reuse the same training batch of data. Increasing the batch size appears to improve performance.

**Expected behavior:**

Running the model on an "old" GPJax training abstraction, with the same model, the same gradient optimiser settings, same number of loop iterations, same random numbers, gives a rough jagged ELBO training curve, as expected. And the fit is excellent, reproducing my previous results.

**Steps to reproduce:**

I'll provide an "old" training loop and or notebook example in a future comment.

thomaspinder:
Resolved in #99

Issue #69, opened 2022-06-02
feat: Getting Rid of TensorFlow


Weizhe-Chen:
# Feature Request

**Describe the Feature Request**

GPJax is a Gaussian processes library based on Jax but currently it still depends on TensorFlow. TensorFlow is annoying because (1) its installation is slow, and (2) it produces many unnecessary warnings when running on CPU-only machines. It would be great if we can implement all functionalities without TensorFlow. Is it possible?

**Describe Preferred Solution**

Replace all the functionalities of TensorFlow with their Jax counterparts.

**If the feature request is approved, would you be willing to submit a PR?**
Yes

Daniel-Dodd:
Hi, thanks for raising this issue. We are aware of this - it would be nice to drop this dependency! We currently rely on `tensorflow.data` to provide data-loading capabilities (prefetching, batching, shuffling, ...) on the `gpjax.Dataset` object for stochastic optimisation. We wouldn't want to fully implement Jax-based versions of these operations in GPJax from scratch (for readability and maintainability). Do you have any suggestions for Jax libraries that provide well-developed data-loading functionality?

thomaspinder:
Hi Weizhe,

Thanks for your suggestion. As Dan has mentioned, removing the TensorFlow dependency is something we'd like to do. Your `elegy` suggestion seems sensible and I'd be very happy to review a PR on this. If you'd like to move forward with this, then there are two things I'd specifically like to confirm 
1. The installation of `elegy` is not larger and/or more convoluted than TensorFlow.
2. Replacing the TensorFlow dataset object with an elegy implementation does not impair speed

Please let me know what, if any, support you will need to get started with this; I'm very happy to assist.

Weizhe-Chen:
Hi Thomas,
Good questions. I had the same concerns before suggesting this.
1. Since we only need to `Dataset` and `DataLoader` classes, it would be better if we directly put these two classes into `GPJax` rather than installing `elegy`. We can remove TF dependency without introducing another heavy burden.
2. It is straightforward to complete the above step. I can compare the speed of `tf.data` and `elegy.data` today and let you know.

thomaspinder:
Responding to 1., this speaks to a broader question I've had as to whether a second library should be formed that provides functionality for just _applying_ GP methods. In that case, GPJax could simply provide the tools that one needs to fit a GP, whilst the second library (let's call it GPFit for now) has implementations of dataloaders, fitting routines, and post-hoc model diagnostics. 

[...]

PR #99, opened 2022-08-19
Remove TensorFlow dependency, Fix fit batches and add history

Daniel-Dodd:
This PR removes TensorFlow (#69  ). Though the mini-batch cacheing bug in the function `fit_batches` (#96 ) can be fixed by JAX's host callback, as per a previous commit on this branch, it is desirable to remove this dependency.

- [ X] Bugfix
- [ X] Feature
- [ X] Code style update (formatting, renaming)
- [ ] Build related changes
- [ X] Documentation content changes

## Other information


gpjax/abstractions.py
```diff
@@ -125,42 +130,56 @@ def fit_batches(
...
         log_rate (int, optional): How frequently the objective function's value should be printed. Defaults to 10.
     Returns:
-        tp.Dict: An optimised set of parameters.
+        tp.Tuple[tp.Dict, f64["n_iters"]]: A tuple comprising optimised parameters and training history respectively.
     """
 
     opt_state = optax_optim.init(params)
-    next_batch = train_data.get_batcher()
+
+    prng = convert_seed(seed)
```
thomaspinder, line 156: Is there a specific reason for this? Why can't we just require the user to supply a Jax PRNGKey?

gpjax/abstractions.py
```diff
@@ -125,42 +130,56 @@ def fit_batches(
...
+        batch = Dataset(X=x[indicies], y=y[indicies])
+
         loss_val, loss_gradient = jax.value_and_grad(loss)(params, batch)
         updates, opt_state = optax_optim.update(loss_gradient, opt_state, params)
         params = optax.apply_updates(params, updates)
-        params_opt_state = params, opt_state
-        return params_opt_state, loss_val
 
-    (params, _), _ = jax.lax.scan(step, (params, opt_state), jnp.arange(n_iters))
+        prng, _ = jr.split(prng)
```
thomaspinder, line 176: Are you sure this is correct? Typically in Jax we'd split the key as `key, subkey = jr.split(key)` and then use subkey for stochastic ops. I believe this is because the return of `jr.split(key, 10) will first return a single key and the seond return is a list of 10 keys for use in a stochastic ops.

gpjax/types.py
```diff
@@ -45,108 +39,6 @@ def out_dim(self) -> int:
...
-        self._make_tfd_if_none()
-        tfd_iter = iter(self._tf_dataset)
-
-        def next_batch() -> "Dataset":
-            X_batch, y_batch = next(tfd_iter)
-
-            return Dataset(X=X_batch.numpy(), y=y_batch.numpy())
-
-        return next_batch
-
```
thomaspinder, line 149: This is very pleasing :)

gpjax/utils.py
```diff
@@ -77,3 +79,14 @@ def array_to_dict(parameter_array) -> tp.Dict:
...
+
+def convert_seed(seed: tp.Union[int, PRNGKey]) -> PRNGKey:
+    """Ensure that seeds type."""
+
+    if isinstance(seed, int):
+        rng = jr.PRNGKey(seed)
+    else:  # key is of type PRNGKey
+        rng = seed
+
+    return rng
```
thomaspinder, line 92: I feel this may be superfluous?

thomaspinder (changes requested):
This PR makes me very happy. I have a few comments that I'd like resolving, then we can merge.

gpjax/abstractions.py
```diff
@@ -2,13 +2,17 @@
 
 import jax
 import jax.numpy as jnp
+import jax.random as jr
 import optax
+from chex import PRNGKey
```
thomaspinder, line 7: Is this intentional? Could we perhaps alias the import as `PRNGKeyType`

gpjax/abstractions.py
Daniel-Dodd: The thinking behind this was to match how things are done in distrax. With fresh eyes, I would agree that this is superfluous. Lets remove it.

gpjax/abstractions.py
Daniel-Dodd: Ooh I did not realise this. I will have to fix this. I think we might want to have the data batch loading code as a separate function so that we can unit test it.

gpjax/types.py
Daniel-Dodd: It was satisfying hitting backspace!

gpjax/utils.py
Daniel-Dodd: Agreed. I'll remove this.

gpjax/abstractions.py
Daniel-Dodd: Good spot! I did not give this much thought, I followed what was done in distrax (see e.g., https://github.com/deepmind/distrax/blob/615605e8701eb55f75854b9b6e13913a9d6d7716/distrax/_src/distributions/distribution.py). Your suggestion of `PRNGKeyType` makes more sense!

Daniel-Dodd:
Thanks for a thorough review @thomaspinder . I'll get on with the changes. Then I can do the same for the natural gradients branch, and then that PR it ready to be reviewed too.

Daniel-Dodd:
@thomaspinder I have resolved the changes.

I have put the batch selection and key updating code in to a separate function. Though less neat, it means we can ensure unit tests assert the correct batch size, dimensions, batch updates are used. Thoughts?

Daniel-Dodd:
@thomaspinder have done a rebase to account for the #100 merge. I need to update the tests before you review.

Daniel-Dodd:
@thomaspinder Ready for review.

gpjax/abstractions.py
```diff
@@ -162,33 +161,44 @@ def fit_batches(
...
 
     @progress_bar_scan(n_iters, log_rate)
     def step(carry, _):
-        params, opt_state, prng = carry
-
-        indicies = jr.choice(prng, n, (batch_size,), replace=True)
+        params, opt_state, current_key = carry
 
-        batch = Dataset(X=x[indicies], y=y[indicies])
+        batch, new_key = get_batch_new_key(train_data, batch_size, current_key)
```
thomaspinder, line 172: I would prefer that the new batch was generated from a specific key i.e., the new key is pased into `get_batch_new_key()`, not the existing key being split within the function. I think for reproducibility, this approach will be cleaner. Unless there's a specific reason for structuring the code as you have done, then I would suggest changing to mimic my suggestion.

gpjax/abstractions.py
```diff
@@ -162,33 +161,44 @@ def fit_batches(
...
+        train_data (Dataset): The training dataset.
+        batch_size (int): The batch size.
+    Returns:
+        Dataset: The batched dataset.
+    """
+    x, y, n = train_data.X, train_data.y, train_data.n
+
+    indicies = jr.choice(key, n, (batch_size,), replace=True)
+
+    _, new_key = jr.split(key)
```
thomaspinder, line 202: Re. comment on ln 172.

gpjax/types.py
```diff
@@ -5,6 +5,8 @@
 
 NoneType = type(None)
 
+from chex import PRNGKey as PRNGKeyType
```
thomaspinder, line 8: Unnecessary given line 2.

thomaspinder (changes requested):
Just two simple changes my side, then this can be merged.

gpjax/types.py
Daniel-Dodd: Oh yes. Good spot.

gpjax/abstractions.py
Daniel-Dodd: Instead of having (in the function `get_batch_new_key`) the order:

```python
  indicies = jr.choice(key, n, (batch_size,), replace=True)

   _, new_key = jr.split(key)
```

You would prefer (?):
```python
   key, new_key = jr.split(key)
   indicies = jr.choice(key, n, (batch_size,), replace=True)
```

Is this what you mean?

Daniel-Dodd: (If so, I'd also prefer this).

thomaspinder: Yes. In fact, to be consistent with Jax documentation, I'd really like 
```python
 key, subkey = jr.split(key)
 indicies = jr.choice(subkey, n, (batch_size,), replace=True)
```

Daniel-Dodd: Fantastic! I'll get this done now.

Daniel-Dodd: Resolved in c31df38

Daniel-Dodd:
> Just two simple changes my side, then this can be merged.

Hi @thomaspinder. These should be resolved in c31df38.

gpjax/abstractions.py
```diff
@@ -162,33 +161,44 @@ def fit_batches(
...
+    """Batch the data into mini-batches.
+    Args:
+        train_data (Dataset): The training dataset.
+        batch_size (int): The batch size.
+    Returns:
+        Dataset: The batched dataset.
+    """
+    x, y, n = train_data.X, train_data.y, train_data.n
+
+    key, new_key = jr.split(key)
```
thomaspinder, line 200: Apologies. Maybe I wasn't very clear. My point was that I don't like the key being split with the function that collects my batch. If I run this function, then I want to be able to pass the function a key and then every time I run the function with the same key I will receive the same data batch. Right now, this won't happen as you're splitting the key within the function. In my opinion, the key should be split __within the training loop__ and then then the new key is passed to this function from which a data batch is returned. The function would then look like:
```python
def get_batch_new_key(
    train_data: Dataset, batch_size: int, key: PRNGKeyType
) -> Dataset:
    """Batch the data into mini-batches.
    Args:
        train_data (Dataset): The training dataset.
        batch_size (int): The batch size.
    Returns:
        Dataset: The batched dataset.
    """
    x, y, n = train_data.X, train_data.y, train_data.n
    indicies = jr.choice(key, n, (batch_size,), replace=True)
    return Dataset(X=x[indicies], y=y[indicies]), new_key
```

gpjax/abstractions.py
thomaspinder, line 172: In accordance with my below comment, this line would become two:
```python
current_key, subkey = jr.split(current_key)
batch = get_batch_new_key(train_data, batch_size, subkey)
```

thomaspinder (approved):

PR #100, opened 2022-08-19
Revamped init

thomaspinder:
Cleans up the `initialise` function such that it accepts a `PRNGKey` object and can overwrite the default parameter values. The return from `fit` and `initialise` now returns a dataclass for code cleanlieness. 

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ x ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes


gpjax/abstractions.py
```diff
@@ -15,6 +15,15 @@
 from .utils import convert_seed
 
 
+@dataclass
```
Daniel-Dodd, line 18: This is a nice idea, keeps everything tidy! 

Just a couple of questions:
- What advantages does having a dataclass have over a named Tuple? 
- Do we want to make ensure the trained params and history immutable after training has finished?

Daniel-Dodd: (Sorry this is in reference to `InferenceState`).

gpjax/abstractions.py
```diff
@@ -120,8 +129,8 @@ def step(carry, i):
         return carry, loss_val
 
     (params, _), history = jax.lax.scan(step, (params, opt_state), jnp.arange(n_iters))
-
-    return params, history
+    inf_state = InferenceState(params=params, history=history)
+    return inf_state
```
Daniel-Dodd, line 133: Nice!

gpjax/likelihoods.py
```diff
@@ -38,6 +37,10 @@ def link_function(self) -> Callable:
         """Return the link function of the likelihood function."""
         raise NotImplementedError
 
+    def init(self, key: jnp.ndarray) -> Dict:
+        """Initialise the stateful parameters of the likelihood function."""
+        return self.params
```
Daniel-Dodd, line 42: What does this do?

(Calling this method results in an error, since `.params` is no longer there).

gpjax/mean_functions.py
```diff
@@ -25,16 +25,26 @@ def __call__(self, x: f64["N D"]) -> f64["N Q"]:
...
+    def init(self, key: jnp.ndarray) -> Dict:
+        """Initialise the stateful parameters of the mean function. This method is required for all subclasses.
+
+        Args:
+            key (Array): The key used to initialise the stateful parameters.
+
+        Returns:
+            dict: The initialised stateful parameters of the mean function.
+        """
+        return self.params
```
Daniel-Dodd, line 46: Like with the likelihoods, I'm not sure what this does given `.params` does not exist.

gpjax/parameters.py
```diff
@@ -1,27 +1,51 @@
...
+class ParameterState:
+    """The state of the model. This includes the parameter set and the functions that allow parameters to be constrained and unconstrained."""
+
+    params: tp.Dict
+    trainables: tp.Dict
+    constrainers: tp.Dict
+    unconstrainers: tp.Dict
+
+    def unpack(self):
+        return self.params, self.trainables, self.constrainers, self.unconstrainers
```
Daniel-Dodd, line 32: I like the neatness of having all of these together. But,  similar to my `InferenceState` comment, what was the thinking behind a `dataclass` instead of something simpler and immutable such as a named tuple.

gpjax/parameters.py
```diff
@@ -1,27 +1,51 @@
...
     constrainers, unconstrainers = build_transforms(params)
     trainables = build_trainables(params)
-    return params, trainables, constrainers, unconstrainers
+    state = ParameterState(
+        params=params,
+        trainables=trainables,
+        constrainers=constrainers,
+        unconstrainers=unconstrainers,
+    )
+    return state
```
Daniel-Dodd, line 48: I like this. Passing a key makes good sense.

Do we have an example of overwriting default parameter values, or initialising them according them to a key? i.e., unless I'm missing something, the method `._initialise_params(key)` seems to default to fixed values, independent of the key for the objects I've looked at thus far.

gpjax/kernels.py
```diff
@@ -59,20 +59,10 @@ def ard(self):
         """
         return True if self.ndims > 1 else False
 
-    @property
-    def params(self) -> dict:
+    def _initialise_params(self, key: jnp.DeviceArray) -> Dict:
         """A template dictionary of the kernel's parameter set."""
         return self._params
```
Daniel-Dodd, line 64: We need to have a think about `._params` on the Kernel object. Firstly, since the `.params` property method no longer exists, we could replace `_params`  with `params` when defining a Kernel.

But the bigger issue is that we should either adopt the same convention for likelihoods and mean functions, or work to remove it. It is important that all objects in GPJax behave the same.

gpjax/kernels.py
```diff
@@ -59,20 +59,10 @@ def ard(self):
...
 
-    @params.setter
-    def params(self, value):
-        """Override the kernel's initial parameter values.
-
-        Args:
-            value (dict): An alternative parameter dictionary that is to be used for initialisation.
-        """
-        self._params = value
-
```
Daniel-Dodd, line 75: This is a relief!

Daniel-Dodd (changes requested):
Hi @thomaspinder, I've done a first pass. 

I very much like how this is developing. This is going to greatly improve GPJax!

Issues that need to be addressed before merging are:

- The `init` method on likelihoods and mean functions are bugs.
- The inconsistencies with kernels to  likelihoods and  mean function regarding `._params`.
- Document what the new `initialise` method is doing and what it returns (i.e., the dataclass object) in e..g, the regression notebook, also document what the training loop returns (i.e., the dataclass object).
- It would be good to have an example with overriding default parameter values on objects, and illustrate the effect of passing a PRNG key to `initialise`.

gpjax/parameters.py
thomaspinder: Resolving as discussion can be held in the earlier comment.

gpjax/parameters.py
thomaspinder: Yes - in latest commit.

gpjax/abstractions.py
thomaspinder: I'm not sure that there's a clear advantage of a dataclass over a tuple. I used a dataclass though for consistency within the rest of the package. Very happy to moved to a namedtuple though if you know of a clear advantage?

Agree on making `InferenceState` immutable - reflected in latest commit.

thomaspinder:
@Daniel-Dodd  - I believe your comments re. updating the notebooks and providing a description for how a parameter's value can be overidden is now present.

Daniel-Dodd:
> @Daniel-Dodd - I believe your comments re. updating the notebooks and providing a description for how a parameter's value can be overidden are now present.

Fantastic, I'll have a read once the docs build is complete.

Daniel-Dodd:
> > @Daniel-Dodd - I believe your comments re. updating the notebooks and providing a description for how a parameter's value can be overidden are now present.
> 
> Fantastic, I'll have a read once the docs build is complete.

@thomaspinder Looks good to me. Can we have also demonstrate using the PRNG key in initialise (and mention what passing it does in the docs)?

thomaspinder:
> > > @Daniel-Dodd - I believe your comments re. updating the notebooks and providing a description for how a parameter's value can be overidden are now present.
> > 
> > 
> > Fantastic, I'll have a read once the docs build is complete.
> 
> @thomaspinder Looks good to me. Can we also demonstrate an example that uses the PRNG key in `initialise` (and mention what passing it does in the docs)?

Right now there is no example where the parameters' initial value is stochastic (do correct me if I'm wrong here), so the key actually does nothing. However, for the RFF/pathwise sampling and the GPLVM in MOGPJax work, it becomes really useful, so I'd suggest deferring that discussion until we have that work merged in.

Daniel-Dodd:
> > > > @Daniel-Dodd - I believe your comments re. updating the notebooks and providing a description for how a parameter's value can be overidden are now present.
> > > 
> > > 
> > > Fantastic, I'll have a read once the docs build is complete.
> > 
> > 
> > @thomaspinder Looks good to me. Can we also demonstrate an example that uses the PRNG key in `initialise` (and mention what passing it does in the docs)?
> 
> Right now there is no example where the parameters' initial value is stochastic (do correct me if I'm wrong here), so the key actually does nothing. However, for the RFF/pathwise sampling and the GPLVM in MOGPJax work, it becomes really useful, so I'd suggest deferring that discussion until we have that work merged in.

Ahhh, okay that makes more sense. 

I thought this was for parameter initialisation through a prior distribution, e.g., to sample initial values for the lengthscale of the kernel.

I would say it is annoying passing a key, if it's not being used for anything (I also think its a bit confusing for new users), but I now understand where you intend to go with it. I would ask if it is possible to have this as a keyword argument or something?

thomaspinder:
> I would say it is annoying passing a key, if it's not being used for anything (I also think its a bit confusing for new users), but I now understand where you intend to go with it. I would ask if it is possible to have this as a keyword argument or something?

It's certainly an option. The only issue is that, due to the use of `**kwargs` within the function, giving a default value to key will result in an error unless the user explicitly writes `key=key` within the function call.

docs/nbs/regression.ipynb
```diff
@@ -154,8 +154,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "params, trainable, constrainer, unconstrainer = gpx.initialise(prior)\n",
-    "prior_dist = prior(params)(xtest)\n",
+    "parameter_state = gpx.initialise(prior, key)\n",
+    "prior_dist = prior(parameter_state.params)(xtest)\n",
```
Daniel-Dodd, line 158: `parameter_state` is clean!

gpjax/kernels.py
```diff
@@ -18,7 +18,6 @@ class Kernel:
     stationary: Optional[bool] = False
     spectral: Optional[bool] = False
     name: Optional[str] = "Kernel"
-    _params: Optional[Dict] = None
```
Daniel-Dodd, line 21: 👍

gpjax/mean_functions.py
```diff
@@ -25,9 +25,8 @@ def __call__(self, x: f64["N D"]) -> f64["N Q"]:
         """
         raise NotImplementedError
 
-    @property
     @abc.abstractmethod
-    def params(self) -> dict:
+    def _initialise_params(self, key: jnp.DeviceArray) -> Dict:
```
Daniel-Dodd, line 29: We can use `PRNGKeyType` in future instead of a DeviceArray on the initialise methods as in #99 .

gpjax/parameters.py
```diff
@@ -1,27 +1,64 @@
...
+        return self.params, self.trainables, self.constrainers, self.unconstrainers
+
+
+def initialise(model, key: jnp.DeviceArray = None, **kwargs) -> ParameterState:
     """Initialise the stateful parameters of any GPJax object. This function also returns the trainability status of each parameter and set of bijectors that allow parameters to be constrained and unconstrained."""
-    params = obj.params
+    if key is None:
+        warn("No PRNGKey specified. Defaulting to seed 123.")
+        key = jr.PRNGKey(123)
+    params = model._initialise_params(key)
```
Daniel-Dodd, line 42: Good. In future, however, we should not raise a warning if the model does not require a key.

Daniel-Dodd (approved):
@thomaspinder happy for this to be merged.

Future issues that need to be addressed:

- We should use `PRNGKeyType` that is used in #99 for PRNG key typing - this is clearer than `jnp.DeviceArray`.
- We should not raise a warning if a model does not require a key in `initialise`.

Issue #105, opened 2022-08-25
dev: Transformations.


Daniel-Dodd:
Perhaps we should not require parameter transformations (via `transform`) in objective functions - they should rest with model training. 

If I have an ELBO or the marginal log-likelihood, shouldn't I just be able to pass my parameters to it without transforming anything?

 For example, for the marginal log-likelihood of GP regression, in the `fit` abstraction we currently have to objective function as `objective = posterior.mll(D, transformation, negative=True)` and `abstractions.py` defines a loss function to train the parameters (that stops gradients):
```python
def loss(params):
        params = trainable_params(params, trainables)
        return objective(params)
```
Perhaps it would be nicer to define an objective with `objective = posterior.mll(D, negative=True)`  (i.e. no transforms specified) and then have the transform in the training loop instead  e.g., 
```python
def loss(params): 
        params = trainable_params(params, trainables)
        params = transform(params, transform)
        return objective(params)
```
The training loop could even possibly take a bijector argument and `abstractions.py` could manage forward and reverse transformations (`gpx.initialise` could return a dictionary of bijectors instead of the `constrainer` and` unconstrainer` convention).

murphyk:
On a related note, it seems that the log-det-jacobian term is missing from the objective function that is passed to HMC.
(At least I could not see the string 'jacfwd' or 'jacrev' anywhere in the codebase :)
The importance of this term is illustrated in https://github.com/probml/pyprobml/blob/master/notebooks/book2/03/change_of_variable_hmc.ipynb

Daniel-Dodd:
Thanks for spotting this @murphyk (and apologies for the slow reply)! We recently refactored `GPJax`, first by removing transformations from all objects in `v0.5.0`. Then, our most recent release `v0.5.2` removes priors from the "marginal likelihood", leaving it as a probability transition kernel only (#153). Priors transformations are now left to the users, and could for example be handled via `transformed.dist` as demonstrated in the [tfp integration notebook example](https://gpjax.readthedocs.io/en/latest/examples/tfp_integration.html).

PR #109, opened 2022-08-31
Transformations

Daniel-Dodd:
The PR refactors transformations #105.

Transformations have been removed from objective functions e.g. the marginal likelihood. So instead of e.g., `objective = posterior.mll(D, transformation, negative=True)`, we now do `objective = posterior.mll(D, negative=True)`. The transformations are instead placed with training loops.

To remove clutter, we remove the dictionary `constrainers` and `unconstrainers` notion, and instead define transformations via bijectors. In particular:

- The `ParameterState` dataclass now comprises `params`,` trainables `and `bijectors`. So e.g. calling `parameter_state = gpx.initialise(posterior, key)`and then `parameter_state.unpack()` returns a tuple `params, trainables, bijectors`.

- The function `gpx.transform` has been removed. To transform parameters to the constrained space, we now do `gpx.constrain(params, bijectors)` and likewise to transform to the unconstrained space, we do `gpx.unconstrain(params, bijectors)`.

Finally, we now define training loops in abstractions e.g. `fit` directly on a `ParameterState` dataclass (i.e., we don't pass through the params dictionary, trainables dictionary and the transformations, as we used to do).

### DIFF
--- a/gpjax/parameters.py
+++ b/gpjax/parameters.py
@@ -22,7 +22,7 @@ Identity = dx.Lambda(forward=lambda x: x, inverse=lambda x: x)
 ################################
 @dataclass
 class ParameterState:
-    """The state of the model. This includes the parameter set and the functions that allow parameters to be constrained and unconstrained."""
+    """The state of the model. This includes the parameter set, which parameters are to be trained and bijectors that allow parameters to be constrained and unconstrained."""
 
     params: tp.Dict
     trainables: tp.Dict
--- a/gpjax/parameters.py
+++ b/gpjax/parameters.py
@@ -139,8 +139,8 @@ def build_bijectors(params: tp.Dict) -> tp.Dict:
     return recursive_bijectors(params, bijectors)
 
 
-def transform(params: tp.Dict, bijectors: tp.Dict, forward: bool) -> tp.Dict:
-    """Transform the parameters according to the constraining or unconstraining function dictionary.
+def constrain(params: tp.Dict, bijectors: tp.Dict) -> tp.Dict:
+    """Transform the parameters to the constrained space for corresponding bijectors.
 
     Args:
         params (tp.Dict): The parameters that are to be transformed.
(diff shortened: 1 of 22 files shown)


gpjax/parameters.py
```diff
@@ -144,48 +139,38 @@ def recursive_bijectors(ps, bs) -> tp.Tuple[tp.Dict, tp.Dict]:
...
-    unconstrainers = jax.tree_util.tree_map(
-        lambda f, b: f(b), unconstrainers, bijectors
-    )
 
-    return constrainers, unconstrainers
-
-
-def transform(params: tp.Dict, transform_map: tp.Dict) -> tp.Dict:
-    """Transform the parameters according to the constraining or unconstraining function dictionary.
+def unconstrain(params: tp.Dict, bijectors: tp.Dict) -> tp.Dict:
```
thomaspinder, line 159: Should these be a function of `params` and `bijectors`, or simply `ParameterState`?

Daniel-Dodd: That's a nice suggestion.

thomaspinder: We could have a utility function call `constrain_state(parameter_state: ParameterState)` and `uncontrain_state(parameter_state: ParameterState)` that have the form:
```
def constrain_state(parameter_state: ParameterState):
    return constrain(parameter_state.params, parameter_state.bijectors)
```

Daniel-Dodd: Could replace this:
```python
def loss(params):
        params = trainable_params(params, trainables)
        params = constrain(params, bijectors)
        return objective(params)
```

with something like?
```python
evaluate(objective, parameter_state)
```

thomaspinder: I guess we need to decide, withing GPJax, do we pass around params as a dictionary, or `ParameterState`.

Daniel-Dodd: Absolutely. Doing things via `ParameterState` is clean and user friendly. It is easy to see the transformations and which parameters are currently trainable. Working with 3 unpacked dictionaries might cause more pain points if you have more than a one model on the go (having parameter state for each one would keep things tidy). The training abstractions would look cleaner. On the other hand working with 3 unpacked dictionaries is more direct with regards to the operations being applied.

Daniel-Dodd:
@thomaspinder The MCMC needs sorting out in classification and TFP notebooks (see built docs). Looking at the old docs, I realise that previously, we didn't transform the parameters to the unconstrained space before passing it to the training loop, and now we do. The trace plots look bad. Moreover, if we do adopt a `ParameterState` convention, it would be good to clean the code up in these.

gpjax/parameters.py
thomaspinder: I wonder whether this is the point at which the package should split - GPJax works directly with the dictionaries and an abstracted package, such as TuneGP, works with the `ParameterState` object to give a more user-friendly interface.

thomaspinder (approved):

thomaspinder:
I've approved everything here. I'm hesistant about merging straight into master though. Can you rebase your natgrads branch with this, or do you require it to be on master?

Issue #115, opened 2022-09-08
feat: Optional training verbosity


thomaspinder:
# Feature Request

Have an argument to control the verbosity of the `fit` method's out.

PR #116, opened 2022-09-20
Add verbose option.

Daniel-Dodd:
Simple implementation of verbosity option #115 . 

Tests don't check if this works or not, but only if any error occurs when turning verbosity on/off.

Training loops tests have been updated to ensure testing of edge case of a single iteration.

### DIFF
--- a/gpjax/abstractions.py
+++ b/gpjax/abstractions.py
@@ -118,6 +118,7 @@ def fit(
     optax_optim: ox.GradientTransformation,
     n_iters: Optional[int] = 100,
     log_rate: Optional[int] = 10,
+    verbose: Optional[bool] = True,
 ) -> InferenceState:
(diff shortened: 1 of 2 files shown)


gpjax/abstractions.py
```diff
@@ -175,6 +178,7 @@ def fit_batches(
     batch_size: int,
     n_iters: Optional[int] = 100,
     log_rate: Optional[int] = 10,
+    verbose: Optional[bool] = True,
```
thomaspinder, line 181: Do you think defaulting to True is sensible? I can see benefits for defaulting to False or True, but the former feels less invasive? Keen to get your thoughts though.

thomaspinder (commented):
My only comment is regarding the verbosity argument's default value. I have no strong feelings but I am curious to hear your thoughts.

PR #125, opened 2022-10-23
Distrax reversion

thomaspinder:
Undo the NumPyro work to revert back to Distrax. This is to ensure more stable parameter transforms.

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ x ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes


theorashid:
Wait, did you implement numpyro and then go back to distrax? Could you explain what happened?

thomaspinder:
Yes - we replaced Distrax/TFP with NumPyro briefly. The main issue was NumPyro's bijectors were giving some odd behaviour. Despite a like-for-like bijector replacement, we were getting nans during optimisation with NumPyro's bijectors. NumPyro also lacked the ability to exploit structured covariance matrices for faster solves/kl-divergences. 

The former issue is perplexing and fundamentally we moved back to Distrax. The latter could have been solved through several PRs but we felt that given our plans to support more structured covariances e.g., kronecker matrices and basis function approximations, implementing our own linear operator and corresponding Gaussian distribution was the most flexible option.

I'd still love to integrate with NumPyro though - their inference suite and broader ML modules is amazing. I know you've experience with NumPyro, so if you have any thoughts on this then I'd be all ears!

theorashid:
Okay interesting. Good to know. I don't have too much experience with the inner working of numpyro for constructing likelihoods etc. When I get nans in my chains, sometimes running `numpyro.enable_x64()` does the job, but I don't know your issues exactly.

I'm more than happy with your approach I saw [dynamax](https://github.com/probml/dynamax/blob/c2352cbcdaef4b24017f810de788716c9a0cdc0f/dynamax/linear_gaussian_ssm/linear_gaussian_ssm.py#L7) had taken a similar approach to you for state-space model, where they use tfp jax distrivutions and use blackjax for sampling. Is there much difference between tfp distributions and distrax?

The main benefit to numpyro is the ability to have a GP as part of a larger probabilistic model. If it's possible to integrate GPJax (or dynamax) into larger probabilistic models built with tfp.jax, that would be great – even if tfp is so much harder to use as an end user than numpyro.

PR #154, opened 2022-11-25
Incorporate JaxLinOp with GPJax

Daniel-Dodd:
This PR incorporates `JaxLinOp` with `GPJax`. 

(Still need to update documentation).


gpjax/gps.py
```diff
@@ -789,7 +778,7 @@ def mll(params: Dict):
             wx = params["latent"]
 
             # f(x) = μx  +  Lx wx
-            fx = μx + jnp.matmul(Lx, wx)
+            fx = μx + Lx @ wx
```
thomaspinder, line 781: I understand why we can't right now but it'd be cleaner if we didn't have to use `@` for matmuls.

gpjax/variational_families.py
```diff
@@ -22,12 +22,15 @@
 from chex import dataclass
 from jaxtyping import Array, Float
 
+from jaxlinop import identity as I
```
thomaspinder, line 25: Same comment as above re. `identitity` and `I`

thomaspinder (changes requested):
Looks good to me. A few minor comments and then we can merge.

gpjax/gps.py
Daniel-Dodd: An issue is open for this on `JaxLinOp` (JaxGaussianProcesses/JaxLinOp/#2).

thomaspinder (approved):

PR #188, opened 2023-02-24
pyspelling

st--:
Here is the pyspelling config, a wordlist (which probably has too many words in), and some more typo fixes. Running `pyspelling` now finishes successfully. Requires [pyspelling](https://github.com/facelessuser/pyspelling) (e.g. `pip install pyspelling`) and `aspell` (Linux package, e.g. `apt-get install aspell`).


.pyspelling.yml
```diff
@@ -0,0 +1,43 @@
...
+      - open: '(?P<open>`+)'
+        close: '(?P=open)'
+      - open: 'http://'
+        close: '[: \n]'
+      - open: 'https://'
+        close: '[: \n]'
+      - open: 'pragma:'
+        close: '$'
+      - open: '__all__ = \['
+        close: ']'
```
st--, line 41: This is supposed to ignore the `__all__ = ["..."]` blocks of modules, but it doesn't quite seem to work yet, that's why there's some extra "words" in the word list that shouldn't have to be there...

.pyspelling.yml
```diff
@@ -0,0 +1,49 @@
+matrix:
+- name: Python Source
+  aspell:
+    lang: en
+    d: en_GB
+  dictionary:
+    wordlists:
+    - wordlist-custom.txt
```
st--, line 8: please suggest a better place/name for putting the wordlist :)

thomaspinder: I'm fine with is being in root.

PR #191, opened 2023-03-16
Correct key handling in examples

frazane:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Documentation content changes

In some of the examples, the same PRNGKey is used to generate both x and the noise in y, and x is sorted to avoid the correlation resulting by using the same key. This could be somewhat confusing for new users (myself included).

## What is the new behavior?

We simply split the key and use two separate keys to generate x and the noise in y.


thomaspinder:
Thanks @frazane - this looks great!

PR #195, opened 2023-03-21
Rff sampler

henrymoss:
I have added a method to the GP prior so that we can get approximate samples from the GP.  I use the standard RFF approach to get samples with functional forms that can be cheaply queried as many times as we like.

### DIFF
--- a/gpjax/gps.py
+++ b/gpjax/gps.py
@@ -34,6 +34,7 @@ from .mean_functions import AbstractMeanFunction, Zero
 from jaxutils import Dataset
 from .utils import concat_dictionaries
 from .gaussian_distribution import GaussianDistribution
+from .types import FunctionalSample
 
 import deprecation
 
@@ -101,13 +102,6 @@ class AbstractPrior(PyTree):
     def _initialise_params(self, key: KeyArray) -> Dict:
         """Deprecated method for initialising the GP's parameters. Succeeded by ``init_params``."""
         return self.init_params(key)
-
-
-FunctionalSample = Callable[[Float[Array, "N D"]], Float[Array, "N B"]]
-""" Type alias for functions representing `B` samples from a model, to be evaluated on any set of
-`N` inputs (of dimension `D`) and returning the evaluations of each (potentially approximate)
-sample draw across these inputs.
-"""
 
 
 #######################
@@ -258,7 +252,7 @@ class Prior(AbstractPrior):
         self,
         num_samples: int,
         params: Dict,
-        seed: KeyArray,
+        key: KeyArray,
         num_features: Optional[int] = 100,
     ) -> FunctionalSample:
         """Build an approximate sample from the Gaussian process prior. This method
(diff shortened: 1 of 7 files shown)


gpjax/gps.py
```diff
@@ -101,6 +103,13 @@ def _initialise_params(self, key: KeyArray) -> Dict:
         return self.init_params(key)
 
 
+FunctionalSample = Callable[[Float[Array, "N D"]], Float[Array, "N B"]]
+""" Type alias for functions representing `B` samples from a model, to be evaluated on any set of
+`N` inputs (of dimension `D`) and returning the evaluations of each (potentially approximate)
+sample draw across these inputs.
+"""
+
+
```
thomaspinder, line 112: Nice. Could this go in a `utils` or `types` file?

gpjax/gps.py
```diff
@@ -247,6 +254,89 @@ def predict_fn(
...
         return predict_fn
 
+    def sample_approx(
+        self,
+        num_samples: int,
+        params: Dict,
+        seed: KeyArray,
+        num_features: Optional[int] = 100,
+    ) -> FunctionalSample:
+        """Build an approximate sample from the Gaussian process prior. This method
```
thomaspinder, line 264: Think you'll need a raw string or double backslashes.

gpjax/gps.py
```diff
@@ -247,6 +254,89 @@ def predict_fn(
...
+            kernel.
+
+
+        Returns:
+            FunctionalSample: A function representing an approximate sample from the Gaussian
+            process prior.
+        """
+        for integer_input in [num_features, num_samples]:
+            if (not isinstance(integer_input, int)) or integer_input < 0:
+                raise ValueError
```
thomaspinder, line 314: Can you be explicit here e.g., `ValueError('input is negative and must be positive')` or `ValueError(f'input is of type {type(integer_input} and must be an integer')`?

thomaspinder (changes requested):
Good PR. Documentation and testing is excellent. Few minor comments to address and then we can merge.

gpjax/gps.py
henrymoss: where?

thomaspinder: `"""Build` should be `r"""Build`

gpjax/gps.py
```diff
@@ -466,12 +552,124 @@ def predict(test_inputs: Float[Array, "N D"]) -> GaussianDistribution:
...
 
+    def sample_approx(
+        self,
+        num_samples: int,
+        params: Dict,
+        train_data: Dataset,
+        key: KeyArray,
+        num_features: Optional[int] = 100,
+    ) -> FunctionalSample:
+        """Build an approximate sample from the Gaussian process posterior. This method
```
thomaspinder, line 567: Same comment as above re. `r"""`

gpjax/gps.py
```diff
@@ -466,12 +552,124 @@ def predict(test_inputs: Float[Array, "N D"]) -> GaussianDistribution:
...
 
         return predict
 
+    def sample_approx(
+        self,
+        num_samples: int,
+        params: Dict,
+        train_data: Dataset,
+        key: KeyArray,
+        num_features: Optional[int] = 100,
```
thomaspinder, line 565: Is 100 sensible here?

thomaspinder (approved):

PR #199, opened 2023-03-27
Kernel refactoring

thomaspinder:
Refactor the kernel module of gpjax to be compatible with PyTree.

## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes


gpjax/parameters/bijectors.py
```diff
@@ -0,0 +1,34 @@
...
+
+    forward: Callable = static_field()
+    inverse: Callable = static_field()
+
+
+Identity = Bijector(forward=lambda x: x, inverse=lambda x: x)
+
+Softplus = Bijector(
+    forward=lambda x: jnp.log(1.0 + jnp.exp(x)),
+    inverse=lambda x: jnp.log(jnp.exp(x) - 1.0),
```
st--, line 33: These are not very numerically stable (and the forward will easily overflow). Better would be
```suggestion
    forward=jax.nn.softplus,
    inverse=tfp.math.softplus_inverse,
```
or, would it be an option to simply use tfp's bijector implementations directly?

gpjax/parameters/module.py
```diff
@@ -0,0 +1,259 @@
...
+
+        Returns:
+            Module: tranformed to the unconstrained space.
+        """
+
+        def _apply_unconstrain(meta_leaf):
+            meta, leaf = meta_leaf
+            return meta.get("bijector", Identity).inverse(leaf)
+
+        return meta_map(_apply_unconstrain, self)
```
st--, line 128: I find this confusing. What's the "base representation"? This makes me think that the bijector is effectively applied twice, i.e. `bijector.forward(bijector.forward(m.unconstrain())) == m.constrain()`, where I would expect the difference between "constrained" and "unconstrained" to be a _single_ application of the bijector :thinking:

gpjax/parameters/bijectors.py
Daniel-Dodd: Nice, thanks @st-- 

Yeah, I favour using TFP. We will be dropping distrax in favour of this too.

(This is just legacy code from Mytree that aimed to be backend dependancy free).

thomaspinder:
Subsumed by PRs into `v0.6`.

PR #229, opened 2023-04-13
fix depreciations & warnings

st--:
Fixes a bunch of deprecated uses internally that were generating warnings.


st--:
Out of date as not targeting v0.6...

PR #230, opened 2023-04-13
Introduce beartype & fix types

### LINKED ISSUE #202: feat: Use beartyping
Use beartyping to type check code.

st--:

- [x] Bugfixes identified by added type checking
- [x] Some backwards-incompatible changes to fix behaviour
- [x] Code style update (formatting, renaming)

No runtime type checking.

Resolves #202 

## What is the new behavior?

Runtime type checking.


st--:
Currently stuck on this issue: https://github.com/beartype/beartype/issues/226#issuecomment-1511019847

st--:
I've resolved the issue https://github.com/beartype/beartype/issues/233 by removing the `from __future__ import annotations` and replacing forward-references with strings instead. Now back to fixing actual type mismatches...

tests/test_kernels/test_approximations.py
```diff
@@ -130,12 +130,12 @@ def test_exactness(kernel):
 )
 def test_value_error(kernel):
     with pytest.raises(ValueError):
-        RFF(kernel(), num_basis_fns=10)
+        RFF(base_kernel=kernel(), num_basis_fns=10)
```
st--, line 133: A better solution would be Python 3.10's KW_ONLY fields for the base classes

tests/test_kernels/test_non_euclidean.py
```diff
@@ -25,6 +26,7 @@
 config.update("jax_enable_x64", True)
 
 
+@pytest.mark.skip("Graph kernel is broken")
```
st--, line 29: Couldn't figure out how to get the shapes/types to match up. Looks like a weird special-case that doesn't quite follow the general kernel API. Please check separately.

tests/test_kernels/test_nonstationary.py
```diff
@@ -64,7 +64,7 @@ def test_initialization(self, fields: dict, dim: int) -> None:
         assert is_dataclass(self.kernel)
 
         # Input fields as JAX arrays
-        fields = {k: jnp.array([v]) for k, v in fields.items()}
+        fields = {k: jnp.array(v) for k, v in fields.items()}
```
st--, line 67: backwards-incompatible change due to using proper scalar fields for e.g. variance

tests/test_kernels/test_stationary.py
```diff
@@ -82,7 +82,7 @@ def test_initialization(self, fields: dict, dim: int) -> None:
         assert is_dataclass(self.kernel)
 
         # Input fields as JAX arrays
-        fields = {k: jnp.array([v]) for k, v in fields.items()}
+        fields = {k: jnp.array(v) for k, v in fields.items()}
```
st--, line 85: backwards-incompatible change due to using proper scalar fields for e.g. variance

tests/test_mean_functions.py
```diff
@@ -13,28 +14,28 @@ def test_abstract() -> None:
...
     assert isinstance(mf, AbstractMeanFunction)
-    assert (mf(jax.numpy.array([1.0])) == constant).all()
-    assert (mf(jax.numpy.array([2.0, 3.0])) == constant).all()
+    assert (mf(jnp.array([[1.0]])) == jnp.array([constant])).all()
+    assert (mf(jnp.array([[2.0, 3.0]])) == jnp.array([constant])).all()
     assert (
-        jax.vmap(mf)(jax.numpy.array([[1.0], [2.0]]))
-        == jax.numpy.array([constant, constant])
+        mf(jnp.array([[1.0], [2.0]]))
+        == jnp.array([constant, constant])
```
st--, line 36: The `Constant` mean function doesn't quite work how you would expect from the base class (simply operate on a single data point, then we vmap over X); instead, it takes in the whole dataset. Should this become an issue to make consistent with kernel's `__call__` later?

tests/test_objectives.py
```diff
@@ -23,15 +23,15 @@ def build_data(num_datapoints: int, num_dims: int, key, binary: bool):
             0.5
             * jnp.sign(
                 jnp.cos(
-                    3 * x[:, 1].reshape(-1, 1)
+                    3 * x[:, 0].reshape(-1, 1)
```
st--, line 26: drive-by fix: the previous code shouldn't have worked for num_dims=1 ... :thinking:

tests/test_variational_families.py
```diff
@@ -44,12 +45,16 @@ def test_abstract_variational_family():
         AbstractVariationalFamily()
 
     # Create a dummy variational family class with abstract methods implemented.
+    class DummyPosterior:
+        @property
+        def __class__(self) -> type: return AbstractPosterior
```
st--, line 50: the right way of mocking for beartype

gpjax/utils.py
st--, line 1: should we rename it gpjax/types.py ?

gpjax/utils.py
```diff
@@ -0,0 +1,34 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+from beartype.typing import Union
+from jaxtyping import Bool, UInt32, Int, Float, Array
+from jax.random import KeyArray as JAXKeyArray
+
+OldKeyArray = UInt32[Array, "2"]
+KeyArray = Union[OldKeyArray, JAXKeyArray]  # for compatibility regardless of enable_custom_prng setting
```
st--, line 21: nb- not tested with the different enable_custom_prng settings

gpjax/kernels/computations/base.py
```diff
@@ -15,15 +15,15 @@
...
+from beartype.typing import Any
 
 from jax import vmap
 from jaxtyping import Array, Float
 
 from gpjax.linops import (DenseLinearOperator, DiagonalLinearOperator,
                           LinearOperator)
 
-Kernel = Any
+Kernel = Any  # TODO: FIX
```
st--, line 26: this doesn't quite seem right. should we just use a string forward reference instead?

gpjax/kernels/__init__.py
```diff
@@ -14,8 +14,9 @@
 # ==============================================================================
 
 """JaxKern."""
-from .approximations import RFF
 from .base import AbstractKernel, ProductKernel, SumKernel
+
+from .approximations import RFF
```
st--, line 19: Has to come _after_ the `.base` imports so we don't get forward-reference issues!

gpjax/gaussian_distribution.py
```diff
@@ -179,11 +181,11 @@ def _sample_n(self, key: KeyArray, n: int) -> Float[Array, "n N"]:
 
         return vmap(affine_transformation)(Z)
 
-    def sample(self,seed: KeyArray, sample_shape: Tuple[int, int]):  # pylint: disable=useless-super-delegation
-      """See `Distribution.sample`."""
-      return self._sample_n(seed, sample_shape[0])
+    def sample(self, seed: KeyArray, sample_shape: Tuple[int, ...]):  # pylint: disable=useless-super-delegation
+        """See `Distribution.sample`."""
+        return self._sample_n(seed, sample_shape[0])  # TODO this looks weird, why ignore the second entry?
```
st--, line 186: Is this a bug that should become its own issue?

gpjax/dataset.py
```diff
@@ -84,7 +83,7 @@ def out_dim(self) -> int:
         return self.y.shape[1]
 
 
-def _check_shape(X: Float[Array, "N D"], y: Float[Array, "N Q"]) -> None:
+def _check_shape(X: Optional[Float[Array, "..."]], y: Optional[Float[Array, "..."]]) -> None:
```
st--, line 86: Here I removed the beartype-shape checking. Could instead simply remove the `_check_shape` function...

gpjax/base/param.py
```diff
@@ -13,12 +13,11 @@
 # limitations under the License.
 # ==============================================================================
 
-from __future__ import annotations
```
st--, line 16: this caused beartype issues (https://github.com/beartype/beartype/issues/233)

gpjax/__init__.py
```diff
@@ -13,19 +13,21 @@
...
-from .mean_functions import Constant, Zero
-from .objectives import (ELBO, CollapsedELBO, ConjugateMLL,
-                         LogPosteriorDensity, NonConjugateMLL)
-from .variational_families import (CollapsedVariationalGaussian,
-                                   ExpectationVariationalGaussian,
-                                   NaturalVariationalGaussian,
-                                   VariationalGaussian,
-                                   WhitenedVariationalGaussian)
+from jaxtyping import install_import_hook
+with install_import_hook("gpjax", "beartype.beartype"):
```
st--, line 17: do we _want_ to push beartype onto end users?

gpjax/base/module.py
```diff
@@ -31,7 +30,9 @@
                               PyTreeCheckpointer, PyTreeCheckpointHandler,
                               RestoreArgs, SaveArgs)
 from simple_pytree import Pytree, static_field
-from typing_extensions import Self
```
st--, line 34: didn't seem to be compatible with beartype

gpjax/fit.py
```diff
@@ -33,7 +33,7 @@
 def fit(
```
st--, line 33: NB: should `fit()` always jit the code itself ? should we make an issue for that?

gpjax/dataset.py
thomaspinder: Is there any reason not to remove `_check_shape` now we use beartype? If not, then let's do it.

gpjax/fit.py
thomaspinder: I think `fit` should just accept a callable. Let's leave it as is for now and, once merged, we can open a PR to resolve #235

gpjax/gaussian_distribution.py
thomaspinder: Yes. Done - #236

gpjax/kernels/base.py
```diff
@@ -32,8 +37,8 @@
 class AbstractKernel(Module):
     """Base kernel class."""
 
-    compute_engine: AbstractKernelComputation = static_field(DenseKernelComputation)
-    active_dims: List[int] = static_field(None)
+    compute_engine: Type[AbstractKernelComputation] = static_field(DenseKernelComputation)
```
thomaspinder, line 40: Why do we need `Type` here?

gpjax/kernels/base.py
```diff
@@ -46,36 +51,36 @@ def cross_covariance(self, x: Float[Array, "N D"], y: Float[Array, "M D"]):
...
 
         Args:
-            x (Float[Array, "N D"]): The matrix or vector that is to be sliced.
+            x (Float[Array, "... D"]): The matrix or vector that is to be sliced.
         Returns:
-            Float[Array, "N S"]: A sliced form of the input matrix.
+            Float[Array, "... Q"]: A sliced form of the input matrix.
         """
-        return x[..., self.active_dims]
+        return x[..., self.active_dims] if self.active_dims is not None else x
```
thomaspinder, line 62: Does this resolve #237?

gpjax/kernels/computations/base.py
thomaspinder: Yes.

gpjax/likelihoods.py
```diff
@@ -146,14 +143,14 @@ def predict(self, dist: tfd.Distribution) -> tfd.Distribution:
         return self.link_function(mean / jnp.sqrt(1.0 + variance))
 
 
-def inv_probit(x: Float[Array, "N 1"]) -> Float[Array, "N 1"]:
+def inv_probit(x: Float[Array, "*N"]) -> Float[Array, "*N"]:
```
thomaspinder, line 146: What does the `*` in `*N` do?

gpjax/linops/dense_linear_operator.py
```diff
@@ -107,7 +103,7 @@ def __mul__(self, other: float) -> LinearOperator:
 
         return DenseLinearOperator(matrix=self.matrix * other)
 
-    def _add_diagonal(self, other: DiagonalLinearOperator) -> LinearOperator:
+    def _add_diagonal(self, other: "gpjax.linops.diagonal_linear_operator.DiagonalLinearOperator") -> LinearOperator:
```
thomaspinder, line 106: Why do we need `gpjax.linops.diagonal_linear_operator` here? Can we not just have `DiagonalLinearOperator`?

gpjax/linops/linear_operator.py
```diff
@@ -35,8 +30,9 @@
 NestedT = Union[T, Iterable["NestedT"], Mapping[Any, "NestedT"]]
 
 # Nested types.
+DTypes = Union[Type[jnp.float32], Type[jnp.float64], Type[jnp.int32], Type[jnp.int64]]
 ShapeT = TypeVar("ShapeT", bound=NestedT[Tuple[int, ...]])
-DTypeT = TypeVar("DTypeT", bound=NestedT[jnp.dtype])
+DTypeT = TypeVar("DTypeT", bound=NestedT[DTypes])
```
thomaspinder, line 35: Could these be moved into `linops.utils` to be consistent with other types?

gpjax/linops/linear_operator.py
```diff
@@ -60,58 +56,58 @@ def ndim(self) -> int:
...
         raise NotImplementedError
 
-    def __rmul__(self, other: float) -> LinearOperator:
+    def __rmul__(self, other: ScalarFloat) -> "LinearOperator":
         """Reimplimentation of multiply linear operator by scalar."""
         return self.__mul__(other)
 
     @abc.abstractmethod
-    def _add_diagonal(self, other: DiagonalLinearOperator) -> LinearOperator:
+    def _add_diagonal(self, other: "gpjax.linops.diagonal_linear_operator.DiagonalLinearOperator") -> "LinearOperator":
```
thomaspinder, line 97: Do we need `gpjax.linops.diagonal_linear_operator`?

gpjax/utils.py
thomaspinder: Yes. Or `typing`?

tests/test_gps.py
```diff
@@ -229,13 +231,13 @@ def test_prior_sample_approx(num_datapoints, kernel, mean_function):
         p.sample_approx(-1, key)
     with pytest.raises(ValueError):
         p.sample_approx(0, key)
-    with pytest.raises(ValueError):
+    with pytest.raises((ValueError, beartype.roar.BeartypeCallHintParamViolation)):
```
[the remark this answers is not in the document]
thomaspinder: I'd prefer to remove `_check` functions? It feels cleaner. Is there a draw back to pushing beartype onto the user?

tests/test_kernels/test_approximations.py
thomaspinder: I would love to depend on 3.10. I worry that it's too restrictive though. Maybe this warrants more discussion.

tests/test_kernels/test_non_euclidean.py
thomaspinder: Could you raise an issue for this? I'm not sure I follow entirely.

tests/test_mean_functions.py
thomaspinder: Yes, the mean functions in general could do with being updated to be consistent with the kernel object.

thomaspinder (changes requested):
Nice work @st--. 

A few minor comments and points of clarification.

gpjax/__init__.py
st--: hm, mainly in some places we might have specified stricter types (array dtypes/shapes) than is strictly required by the code, so some code might have run fine if only beartype hadn't intervened. also, the explicit _check_shape error messages might be a bit more informative than the generic beartype ones. I don't know if there's also some more interaction with the jaxtyping `@jaxtyped` decorator.

gpjax/likelihoods.py
st--: like in `*args`, it gobbles up arbitrary dimensions. but beartype itself doesn't check the two are consistent

gpjax/linops/dense_linear_operator.py
st--: no, because it's in a different module, so you need to have the fully-qualified module name for beartype to be able to find the reference

gpjax/linops/linear_operator.py
st--: it's not linops-specific so I would actually favour moving it into something like gpjax.types

gpjax/linops/linear_operator.py
st--: yes, because it's a reference to a class in a different module

gpjax/kernels/base.py
st--: Because it's a class, not an instance.  The type of `DenseKernelComputation` is `Type[DenseKernelComputation]`. Compare to `DenseKernelComputation()`, whose type is `DenseKernelComputation`.

gpjax/dataset.py
st--: Actually, beartype only checks arguments individually, and does not check consistency of dimensions across multiple arguments/return values, so might be better to keep the `_check_shape` (so e.g. "N1 D" and "N2 D" shapes get flagged as not-matching).

gpjax/utils.py
st--: "types" feels more natural to me in principle, but it's stdlib `typing` and `jaxtyping`, so let's go with that.

tests/test_kernels/test_non_euclidean.py
st--: Should've been more specific; I don't think the kernel itself is necessarily broken. Just the shape annotations really didn't match the actual use, and I couldn't (didn't want to) get my head around it and decided to just skip the test instead.

Maybe the thing to do instead, for now, is simply remove all the shape annotations on GraphKernel ?

tests/test_gps.py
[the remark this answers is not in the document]
st--: See https://github.com/JaxGaussianProcesses/GPJax/pull/230#discussion_r1176047677

tests/test_mean_functions.py
st--: opened #238

gpjax/utils.py
st--: done

gpjax/kernels/computations/base.py
st--: done

gpjax/kernels/base.py
st--: Don't know, do we have a MRE ready?

gpjax/dataset.py
st--: Actually actually, jaxtyping itself checks those, I just had a bug in my toy example trying it out. But the error messages are still more verbose/less precise than the ones emitted by _check_shape :disappointed:

tests/test_kernels/test_non_euclidean.py
thomaspinder: Understood. I'm fine with removing the shape annotations for now.

st--:
Going through the notebooks, another issue: `Float[Array, "..."]` specifically matches JAX-arrays, but not numpy arrays. How do JAX-functions handle numpy-arrays? Do they implicitly convert them? Should we therefore relax type hints on all user-facing parts (not just Dataset but also e.g. predict()) to allow numpy arrays?

gpjax/dataset.py
thomaspinder: OK. I'm guessing there's no easy way for us to customise the error messages that Beartype throws?

thomaspinder: If the answer to my above question is no, then let's keep `_check_shape`.

thomaspinder: It looks like this is a *very* hot topic right now: https://github.com/google/jaxtyping/issues/6.

gpjax/__init__.py
thomaspinder: So if we didn't push beartype onto the end-user, then it would just be a testing utility for the package? I could easily imagine how any overly rigid beartype assertions could be annoying for an end user.

thomaspinder:
> Going through the notebooks, another issue: `Float[Array, "..."]` specifically matches JAX-arrays, but not numpy arrays. How do JAX-functions handle numpy-arrays? Do they implicitly convert them? Should we therefore relax type hints on all user-facing parts (not just Dataset but also e.g. predict()) to allow numpy arrays?

Yes, JAX will implicitly convert them. So calling `posterior(np_array, D)` will produce the same result as `posterior(jnp.asarray(np_array), D)`. I think it'd place less friction on the user if we relax the type hint to accomodate Numpy arrays.

tests/test_kernels/test_non_euclidean.py
st--: i've actually managed to fix the annotations to what's happening, but it still seems wrong

gpjax/__init__.py
st--: Yeah! Though on the other hand, it could help us discover more bugs not handled by the tests...

tests/test_kernels/test_non_euclidean.py
st--: See #239

gpjax/dataset.py
```diff
@@ -31,8 +30,8 @@ class Dataset(Pytree):
         y (Optional[Float[Array, "N Q"]]): Output data.
     """
 
-    X: Optional[Float[Array, "N D"]] = None
-    y: Optional[Float[Array, "N Q"]] = None
+    X: Optional[Union[Num[Array, "N D"], Num[Array, "..."]]] = None
+    y: Optional[Union[Float[Array, "N Q"], Float[Array, "..."]]] = None
```
st--, line 34: should this also be
```suggestion
    y: Optional[Union[Num[Array, "N Q"], Num[Array, "..."]]] = None
```
for e.g. classification?

examples/graph_kernels.pct.py
```diff
@@ -85,9 +85,9 @@
 
 true_kernel = gpx.GraphKernel(
     laplacian=L,
-    lengthscale=jnp.array([2.3]),
-    variance=jnp.array([3.2]),
-    smoothness=jnp.array([6.1]),
+    lengthscale=jnp.array(2.3),
+    variance=jnp.array(3.2),
+    smoothness=jnp.array(6.1),
```
st--, line 90: NB- should it be possible as a user to simply pass numpy floats, e.g.
```suggestion
    lengthscale=2.3,
    variance=3.2,
    smoothness=6.1,
```
?

gpjax/__init__.py
st--: For example, a bunch of them were only uncovered by the notebooks. But I guess we could add beartype there as well, and then use that as a way to suggest to users that they _should_ use e.g. beartype (and report when one of our type hints is wrong)?

gpjax/kernels/nonstationary/linear.py
```diff
@@ -17,7 +17,9 @@
 
 import jax.numpy as jnp
 import tensorflow_probability.substrates.jax.bijectors as tfb
-from jaxtyping import Array, Float
+from gpjax.typing import Array
+from jaxtyping import Float
+from gpjax.typing import ScalarFloat
```
thomaspinder, line 22: Can we merge lines 20 and 22 please? I see this occurs for most of the kernel modules.

gpjax/linops/linear_operator.py
thomaspinder: Sure, let's do that.

gpjax/__init__.py
thomaspinder: That's a good way to position it. Let's push it onto users then. If it becomes an issue, we can always walk it back with little major disruption.

thomaspinder (approved):
All looks good to me. There's a small issue with imports, but I'm happy to merge as I've setup a pre-commit action on `docs_update` that will fix this. I will rebase that branch and correct.

If you're happy with the proposed approach of pushing beartype onto the user, then we can merge this.

gpjax/kernels/nonstationary/linear.py
st--: Yeah, that's because I didn't want to touch two dozen files by hand and used `sed` for fixing the imports. I would prefer _not_ to merge those lines manually (a lot more effort because it's not always exactly the same so can't as easily write a `sed` line for it), instead, I would suggest tools like `isort` for consistent import order (similar to `black` for code formatting). But I would leave that for a separate PR that does nothing else other than "visual" changes (so e.g. one can ignore it in git blame).

thomaspinder: Yep, on `docs_update` that included in the pre-commit, so let's leave that job for then.

gpjax/__init__.py
st--: Umm, do you mean _force_ it on users (by including it in the general import), or _strongly suggest it_ to users by having it at the start of every notebook (and inside the tests)? I think I'd be in favour of the latter actually...

thomaspinder: The latter - it's consistent with how we encourage people to use float64 without enforcing it.

st--: done

gpjax/dataset.py
```diff
@@ -31,8 +30,8 @@ class Dataset(Pytree):
         y (Optional[Float[Array, "N Q"]]): Output data.
     """
 
-    X: Optional[Float[Array, "N D"]] = None
-    y: Optional[Float[Array, "N Q"]] = None
+    X: Optional[Union[Num[Array, "N D"], Num[Array, "..."]]] = None
+    y: Optional[Union[Num[Array, "N Q"], Num[Array, "..."]]] = None
```
st--, line 34: I don't remember why I put it in but the dataset fields should be consistent and always be "N D" / "N Q" even if D or Q are 1?

st--: on second thought it might have been so that it actually gets to the `_check_shape` in the post_init (otherwise beartype errors at the constructor already). maybe let's just leave it that way:
```suggestion
    X: Optional[Num[Array, "N D"]] = None
    y: Optional[Num[Array, "N Q"]] = None
```

st--: does it make sense that they're optional, actually?

gpjax/dataset.py
```diff
@@ -84,7 +83,7 @@ def out_dim(self) -> int:
         return self.y.shape[1]
 
 
-def _check_shape(X: Float[Array, "N D"], y: Float[Array, "N Q"]) -> None:
+def _check_shape(X: Optional[Num[Array, "..."]], y: Optional[Float[Array, "..."]]) -> None:
```
st--, line 86:
```suggestion
def _check_shape(X: Optional[Num[Array, "..."]], y: Optional[Num[Array, "..."]]) -> None:
```

st--:
@thomaspinder ok I think it's ready to merge now. I hope I didn't accidentally break anything along the way :sweat_smile:

PR #245, opened 2023-04-30
Arccosine kernel

henrymoss:
I have implemented the ArcCosine kernel for orders 0, 1 and 2. My implementation closely follows that of GPFlow, i.e. it includes some additional flexibility by specifying weight and bias variances.

I have added some extra tests, including one cool Monte carlo one.

Let me know if you can think of more tests?  

I have yet to try it in a GP. Do we do this in the tests anywhere ? (i.e. try all our kernels in GPs)

### DIFF
--- a/tests/test_kernels/test_nonstationary.py
+++ b/tests/test_kernels/test_nonstationary.py
@@ -187,3 +187,32 @@ class TestArcCosine(BaseTestKernel):
     def test_defaults(self, order: int) -> None:
         with pytest.raises(ValueError):
             self.kernel(order=order)
+
+    @pytest.mark.parametrize("order", [0, 1, 2], ids=lambda x: f"order={x}")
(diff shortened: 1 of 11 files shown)


tests/test_kernels/test_nonstationary.py
```diff
@@ -101,7 +96,9 @@ def test_initialization(self, fields: dict, dim: int) -> None:
         # Check meta leaves
         meta = kernel._pytree__meta
         assert not any(f in meta for f in self.static_fields)
-        assert list(meta.keys()) == sorted(set(fields) - set(self.static_fields))
+        assert sorted(list(meta.keys())) == sorted(
```
henrymoss, line 99: Note that this is a bug fix!

henrymoss: Its triggered as I have a bias_variance param which sits near the start of the alphabet

thomaspinder (approved):

PR #246, opened 2023-04-30
Add static field to base, update tests to run on 3.11, fix 3.11 compatibility

Daniel-Dodd:
This PR comprises the following changes:

- (1) Add `static_field` to base, and use this in the testing.

- (2) Limit `simple_pytree` imports to 3 places (and the tests only):
    - (a)`gpjax.base`
    - (b)`gpjax.linops.linear_operator` (We should update this to a module in future).
    - (c)`gpjax.dataset` (We should update this to a module in future).

There should be no other imports. 

- Will need to remove `simple_pytree` imports from the `docs_update` branch in a separate PR.
- Perhaps ought to consider (in a future PR), only using `Module` and limiting `simple_pytree` import to `gpjax.base`.

This PR  also fixes compatibility issues with Python 3.11, that requires dataclass fields to be hashable (while jax.Arrays are not...).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ jobs:
       matrix:
         # Select the Python versions to test against
         os: ["ubuntu-latest", "macos-latest"]
-        python-version: ["3.8", "3.9", "3.10"]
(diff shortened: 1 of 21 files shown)


frazane (approved):
Looks good to me!

PR #258, opened 2023-05-12
Update gpu guide to point to JAX installation guide.

daniel-dodd:


docs/installation.md
```diff
@@ -14,25 +14,10 @@ pip install gpjax
...
     ```
 
-## GPU support
+## GPU/TPU support
 
-GPU support is enabled through proper configuration of the underlying
-[Jax](https://github.com/google/jax) installation. CPU enabled forms of
-both packages are installed as part of the `GPJax` installation. For GPU
-Jax support, the following commands should be run:
+Fancy using GPJax on GPU/TPU? Then you'll need to install JAX with the relevant hardware acceleration support as detailed in these instructions(https://github.com/google/jax#installation).
```
thomaspinder, line 19:
```suggestion
Fancy using GPJax on GPU/TPU? Then you'll need to install JAX with the relevant 
hardware acceleration support as detailed in the 
[JAX installation guide](https://github.com/google/jax#installation).
```

Try and keep lines in md files to 88 characters or less (⌥+q) in VSCode. It's also best to use markdown hyperlinks, not raw urls.

thomaspinder (approved):

PR #271, opened 2023-05-18
Add isotropy test

thomaspinder:
## Type of changes

- [X] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

In GPJax versions <0.6, we have a bug that prevented users defining an isotropic kernel when their data was greater than a single dimension. This issue was inadvertently fixed in the v0.6 release. This PR simply adds a test to make sure an isotropic kernel can be specified.

Issue Number: #237

### DIFF
--- a/tests/test_kernels/test_stationary.py
+++ b/tests/test_kernels/test_stationary.py
@@ -162,6 +162,13 @@ class BaseTestKernel:
             assert sdensity.loc == jnp.array(0.0)
             assert sdensity.scale == jnp.array(1.0)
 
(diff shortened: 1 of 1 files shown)


daniel-dodd:
@thomaspinder, this bug was fixed because now we pass a lengthscale upon defining a class that by default this is a single dimension, giving isotropy?

What happens if the active dims and the lengthscale are different sizes? Does this potential mismatch cause a new bug?

daniel-dodd (approved):

Issue #295, opened 2023-06-08
docs: Likelihoods notebook


thomaspinder:
Provide a notebook that sheds some light on our likelihood construction and what goes on under the hood.

Issue #294, opened 2023-06-07
docs: Integrator example.


daniel-dodd:
Add short example of choosing an integrator being used to the svgp notebook #283.

PR #296, opened 2023-06-08
Add integrator/likelihoods doc

thomaspinder:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [X] Documentation / docstrings
- [ ] Tests
- [ ] Other

## Checklist

## Description

This PR introduces a piece of documentation that provides users with a guide to likelihood functions and how the integration strategy attached to a likelihood function can be changed.

Issue Number: #295 #294

### DIFF
--- a/docs/examples/likelihoods_guide.py
+++ b/docs/examples/likelihoods_guide.py
@@ -194,7 +194,7 @@ for ax in axes.ravel():
 # distribution.
 #
 # A table of commonly used link functions and their corresponding likelihood can be
-# found [here](https://en.wikipedia.org/wiki/Generalized_linear_model#Link_function)
+# found [here](https://en.wikipedia.org/wiki/Generalized_linear_model#Link_function).
 #
 # ### Expected log likelihood
 #
--- a/docs/examples/likelihoods_guide.py
+++ b/docs/examples/likelihoods_guide.py
@@ -86,7 +86,7 @@ ax.legend()
 # likelihood is appropriate for this dataset as it allows for negative values.
 #
 # As we see in \eqref{eq:likelihood_fn}, the likelihood function factorises over the
-# $N$ observations. As such, we must provide this information to GPJax when
+# $n$ observations. As such, we must provide this information to GPJax when
 # instantiating a likelihood object. We do this by specifying the `num_datapoints`
 # argument.
 
--- a/docs/examples/likelihoods_guide.py
+++ b/docs/examples/likelihoods_guide.py
@@ -181,7 +181,7 @@ for ax in axes.ravel():
 
 # ### Link functions
 #
-# In the above figure we can see the latent samples being constrained to be either 0 or
+# In the above figure, we can see the latent samples being constrained to be either 0 or
 # 1 when a Bernoulli likelihood is specified. This is achieved by the
 # `inverse link_function` $\eta(\cdot)$ of the likelihood. The link function is a
 # deterministic function that maps the latent distribution of the Gaussian process to
--- a/docs/examples/likelihoods_guide.py
+++ b/docs/examples/likelihoods_guide.py
@@ -45,7 +45,7 @@
 # In GPJax, all likelihoods are a subclass of the `AbstractLikelihood` class. This base
 # abstract class contains the three core methods that all likelihoods must implement:
 # `predict`, `link_function`, and `expected_log_likelihood`. We will discuss each of
-# these methods in the forthcoming sections, but first we will show how to instantiate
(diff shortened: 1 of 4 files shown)


docs/examples/likelihoods_guide.py
```diff
@@ -0,0 +1,263 @@
+# # Likelihood guide
+#
+# In this notebook we will walk users through the process of creating a new likelihood
+# in GPJax.
```
daniel-dodd, line 4:
```suggestion
# In this notebook, we will walk users through the process of creating a new likelihood
# in GPJax.
```

docs/examples/likelihoods_guide.py
```diff
@@ -0,0 +1,263 @@
...
+# case, we would choose a likelihood function that is only defined for values between
+# 0 and 1.
+#
+# ## Likelihoods in GPJax
+#
+# In GPJax, all likelihoods are a subclass of the `AbstractLikelihood` class. This base
+# abstract class contains the three core methods that all likelihoods must implement:
+# `predict`, `link_function`, and `expected_log_likelihood`. We will discuss each of
+# these methods in the forthcoming sections, but first we will show how to instantiate
+# a likelihood object. To do this, we'll need a dataset.
```
daniel-dodd, line 49:
```suggestion
# these methods in the forthcoming sections, but first, we will show how to instantiate
# a likelihood object. To do this, we'll need a dataset.
```

docs/examples/likelihoods_guide.py
```diff
@@ -0,0 +1,263 @@
...
+import tensorflow_probability.substrates.jax as tfp
+
+tfd = tfp.distributions
+plt.style.use("./gpjax.mplstyle")
+cols = plt.rcParams["axes.prop_cycle"].by_key()["color"]
+key = jr.PRNGKey(123)
+
+
+N = 50
+x = jnp.sort(jr.uniform(key=key, shape=(N, 1), minval=-3.0, maxval=3.0), axis=0)
```
daniel-dodd, line 71: Consistency -> lower case `n` is used in the other notebooks.
```suggestion
n = 50
x = jnp.sort(jr.uniform(key=key, shape=(n, 1), minval=-3.0, maxval=3.0), axis=0)
```

docs/examples/likelihoods_guide.py
```diff
@@ -0,0 +1,263 @@
...
+ax.legend()
+# -
+
+# In this example, our observations have support $[-3, 3]$ and are generated from a
+# sinusoidal function with Gaussian noise. As such, our response values $\mathbf{y}$
+# range between $-1$ and $1$, subject to Gaussian noise. Due to this, a Gaussian
+# likelihood is appropriate for this dataset as it allows for negative values.
+#
+# As we see in \eqref{eq:likelihood_fn}, the likelihood function factorises over the
+# $N$ observations. As such, we must provide this information to GPJax when
```
daniel-dodd, line 89: A lower case $n$ is used through out other sections of the docs.
```suggestion
# As we see in \eqref{eq:likelihood_fn}, the likelihood function factorises over the
# $n$ observations. As such, we must provide this information to GPJax when
```

docs/examples/likelihoods_guide.py
```diff
@@ -0,0 +1,263 @@
...
+        alpha=0.5,
+        color=cols[1],
+        label="Predictive samples",
+    )
+# -
+
+# ### Link functions
+#
+# In the above figure we can see the latent samples being constrained to be either 0 or
+# 1 when a Bernoulli likelihood is specified. This is achieved by the
```
daniel-dodd, line 185:
```suggestion
# In the above figure, we can see the latent samples being constrained to be either 0 or
# 1 when a Bernoulli likelihood is specified. This is achieved by the
```

daniel-dodd (commented):
### To merge:
Looks good @thomaspinder. Left some comments.


### For future:

Its nice that you break down the `AbstractLikelihood` down. As a further suggestion, it would be nice if we explicitly demonstrated what a user would have to code themselves if they wanted to add their own likelihood (like we do in the kernels guide), as an example. E.g., via a Poisson or Bernoulli (or Both). You could even maybe have a table that breaks things down with the maths and the code for the link functions, what the data looks like under each scenario -> e.g., count use Poisson, for Binary use Bernoulli, etc.

And it would be good to explicitly have links/ advertise custom likelihood, kernel, mean function etc, somewhere in our docs.

thomaspinder:
Thanks @daniel-dodd - all suggestions actioned.

Issue #217, opened 2023-01-25
dev: Add paper citations to objects


thomaspinder:
Attach the citation to objects where there is a clear citation available e.g., Rahimi & Recht (2009) for random Fourier features. One option could be to return the bibtex string, or simply the paper name and authors.

PR #300, opened 2023-06-09
Add cite functionality

thomaspinder:
## Type of changes

- [ ] Bug fix
- [X] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

This PR introduces a citation function that, given an object, returns the relevant bibtex string.

Issue Number: #217

### DIFF
--- a/docs/examples/graph_kernels.py
+++ b/docs/examples/graph_kernels.py
@@ -134,9 +134,23 @@ cbar = plt.colorbar(sm)
 
 # %%
 likelihood = gpx.Gaussian(num_datapoints=D.n)
-prior = gpx.Prior(mean_function=gpx.Zero(), kernel=gpx.GraphKernel(laplacian=L))
+kernel = gpx.GraphKernel(laplacian=L)
+prior = gpx.Prior(mean_function=gpx.Zero(), kernel=kernel)
 posterior = prior * likelihood
 
+# %% [markdown]
+#
+# For researchers and the curious reader, GPJax provides the ability to print the
+# bibtex citation for objects such as the graph kernel through the `cite()` function.
+
+# %%
+print(gpx.cite(kernel))
+
+# %% [markdown]
+#
+# With a posterior defined, we can now optimise the model's hyperparameters.
+
+# %%
 opt_posterior, training_history = gpx.fit(
     model=posterior,
     objective=jit(gpx.ConjugateMLL(negative=True)),
--- a/docs/examples/regression.py
+++ b/docs/examples/regression.py
@@ -179,7 +179,7 @@ posterior = prior * likelihood
 # these parameters by optimising the marginal log-likelihood (MLL).
 
 # %%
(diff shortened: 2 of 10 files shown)


docs/examples/graph_kernels.py
```diff
@@ -134,9 +134,23 @@
...
 posterior = prior * likelihood
 
+# %% [markdown]
+#
+# For researchers and the curious reader, GPJax provides the ability to print the
+# bibtex citation for objects such as the graph kernel through the `cite()` function.
+
+# %%
+print(gpx.cite(kernel))
+
```
daniel-dodd, line 148: I have a couple of questions on citations, and am using this one as an example.

calling the cite function on the graph kernel gives:
```
 @inproceedings{borovitskiy2021matern,
authors = {Borovitskiy, Viacheslav and Azangulov, Iskander and Terenin, Alexander and Mostowsky, Peter and Deisenroth, Marc and Durrande, Nicolas},
title = {Matérn Gaussian Processes on Graphs},
year = {2021},
booktitle = {International Conference on Artificial Intelligence and Statistics},
}
```

My questions are:

- (1) `Gaussian` in `title` should be capitalised, but this won't necessarily be capitalised in Latex by default, is capitalisation something we want to enforce for all the citations?
- (2) are things like the `é`  in `Matérn` in `title` latex friendly?

docs/examples/regression.py
```diff
@@ -179,7 +179,7 @@
 # these parameters by optimising the marginal log-likelihood (MLL).
 
 # %%
-negative_mll = jit(gpx.objectives.ConjugateMLL(negative=True))
+negative_mll = gpx.objectives.ConjugateMLL(negative=True)
 negative_mll(posterior, train_data=D)
```
daniel-dodd, line 183: So, I am guessing the cite function as it stands does not work on traced objects? 

- (1) Is it possible to get it to work here in future? And if so open an issue for this.
- (2) If not, what does `cite` currently return? Can we check if the function is traced, and return a helpful error e.g., `TypeError('cite does not work on jit traced objects, please use non-jitted one')`?

We could build either (1) or (2) in to the `dispatch` decorator.

daniel-dodd (commented):
Looks neat @thomaspinder, just left some questions.

docs/examples/graph_kernels.py
thomaspinder: 1) I would always recommend people to have `\DeclareFieldFormat{titlecase}{#1}` in their preamble to spare having to have lots of {} pairs in your bib file as they make the reading significantly messier. I can add {} if you feel strongly that we need them here.
2) This is fine in my experience.

docs/examples/regression.py
thomaspinder: I can open an issue for (1) - it may be possible, but I don't know how :/ I'll do your suggestion for 2 in the revision.

daniel-dodd (approved):

Issue #297, opened 2023-06-08
feat: Implement beta likelihood


thomaspinder:
# Feature Request

Implement the Beta likelihood function. This can be easily achieved as the link function and integration strategy will be identical to the Bernoulli likelihood.

**Related Code**

https://github.com/JaxGaussianProcesses/GPJax/blob/main/gpjax/likelihoods.py#L170

brandall25:
@thomaspinder can I have a go at this?

thomaspinder:
Of course! Do let me know if you need a hand.

brandall25:
Hi @thomaspinder , this is what I have so far:

```python
@dataclass
class Beta(AbstractLikelihood):
    def link_function(self, alpha: float, beta: float) -> tfd.Distribution:
        r"""The probit link function of the Beta likelihood.

        Args:
            alpha : alpha parameter, indicating mean number of successes
            beta  : beta parameter, indicating mean number of failures

        Returns
        -------
            tfd.Distribution: The likelihood function.
        """
        return tfd.Beta(tf.constant(alpha), tf.constant(beta))

    def predict(self, dist: tfd.Distribution, alpha: float, beta: float) -> tfd.Distribution:
        r"""Evaluate the pointwise predictive distribution.

        Evaluate the pointwise predictive distribution, given a Gaussian
        process posterior and likelihood parameters.

        Args:
            dist (tfd.Distribution): The Gaussian process posterior, evaluated
                at a finite set of test points.
            alpha (float) : alpha parameter, indicating mean number of successes
            beta  (float) : beta parameter, indicating mean number of failures

        Returns
        -------
            tfd.Distribution: The pointwise predictive distribution.
        """
        return self.link_function(alpha, beta)
```

Couple questions:

1. Should I keep alpha and beta as inputs to the predict function? Or, should I try to infer them from the base distribution class?
2. Do I need to specify any of the other arguments for the call to tfd.Beta()?

Thanks!

thomaspinder:
Hi @brandall25. This is starting to look good! Are you able to open this as a PR to main? It'll be easier for me to make comments and give feedback there.

PR #301, opened 2023-06-09
Added beta likelihood function

brandall25:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Added beta likelihood function to `likelihoods.py`

Issue Number: #297

### DIFF
--- a/gpjax/likelihoods.py
+++ b/gpjax/likelihoods.py
@@ -201,22 +201,20 @@ class Bernoulli(AbstractLikelihood):
 
 @dataclass
 class Beta(AbstractLikelihood):
-    def link_function(self, alpha: float, beta: float) -> tfd.Distribution:
+    def link_function(self, f: Float[Array, "..."]) -> tfd.Distribution:
         r"""The probit link function of the Beta likelihood.
 
         Args:
-            alpha : alpha parameter, indicating mean number of successes
-            beta  : beta parameter, indicating mean number of failures
+            f (Float[Array, "..."]): Function values.
 
         Returns
         -------
             tfd.Distribution: The likelihood function.
         """
-        return tfd.Beta(alpha, beta)
 
-    def predict(
-        self, dist: tfd.Distribution, alpha: float, beta: float
-    ) -> tfd.Distribution:
+        return tfd.Beta(f[0], f[1])
+
+    def predict(self, dist: tfd.Distribution) -> tfd.Distribution:
         r"""Evaluate the pointwise predictive distribution.
 
         Evaluate the pointwise predictive distribution, given a Gaussian
@@ -225,14 +223,18 @@ class Beta(AbstractLikelihood):
         Args:
             dist (tfd.Distribution): The Gaussian process posterior, evaluated
                 at a finite set of test points.
-            alpha (float) : alpha parameter, indicating mean number of successes
-            beta  (float) : beta parameter, indicating mean number of failures
 
         Returns
         -------
             tfd.Distribution: The pointwise predictive distribution.
         """
-        return self.link_function(alpha, beta)
+        mean = dist.mean().ravel()
+        stddev = dist.stddev().ravel()
+
+        alpha = mean / stddev
+        beta = (1.0 - mean) / stddev
+
+        return self.link_function([alpha, beta])
 
 
 @dataclass
--- a/gpjax/likelihoods.py
+++ b/gpjax/likelihoods.py
@@ -212,7 +212,7 @@ class Beta(AbstractLikelihood):
             tfd.Distribution: The likelihood function.
         """
 
-        return tfd.Beta(f[0], f[1])
+        return tfd.Beta(inv_probit(f[0]), inv_probit(f[1]))
 
     def predict(self, dist: tfd.Distribution) -> tfd.Distribution:
         r"""Evaluate the pointwise predictive distribution.
--- a/gpjax/likelihoods.py
+++ b/gpjax/likelihoods.py
@@ -200,6 +200,42 @@ class Bernoulli(AbstractLikelihood):
 
 
 @dataclass
+class Beta(AbstractLikelihood):
+    def link_function(self, alpha: float, beta: float) -> tfd.Distribution:
+        r"""The probit link function of the Beta likelihood.
+
+        Args:
+            alpha : alpha parameter, indicating mean number of successes
+            beta  : beta parameter, indicating mean number of failures
+
+        Returns
+        -------
+            tfd.Distribution: The likelihood function.
+        """
+        return tfd.Beta(alpha, beta)
+
+    def predict(
+        self, dist: tfd.Distribution, alpha: float, beta: float
+    ) -> tfd.Distribution:
+        r"""Evaluate the pointwise predictive distribution.
+
+        Evaluate the pointwise predictive distribution, given a Gaussian
+        process posterior and likelihood parameters.
+
+        Args:
+            dist (tfd.Distribution): The Gaussian process posterior, evaluated
+                at a finite set of test points.
+            alpha (float) : alpha parameter, indicating mean number of successes
+            beta  (float) : beta parameter, indicating mean number of failures
+
+        Returns
+        -------
+            tfd.Distribution: The pointwise predictive distribution.
+        """
+        return self.link_function(alpha, beta)
+
+
+@dataclass
 class Poisson(AbstractLikelihood):
     def link_function(self, f: Float[Array, "..."]) -> tfd.Distribution:
         r"""The link function of the Poisson likelihood.
@@ -249,6 +285,7 @@ __all__ = [
     "NonGaussianLikelihood",
     "Gaussian",
     "Bernoulli",
+    "Beta",
     "Poisson",
     "inv_probit",
 ]


gpjax/likelihoods.py
```diff
@@ -199,6 +199,42 @@ def predict(self, dist: tfd.Distribution) -> tfd.Distribution:
         return self.link_function(mean / jnp.sqrt(1.0 + variance))
 
 
+@dataclass
+class Beta(AbstractLikelihood):
```
thomaspinder, line 203: You'll need to specify the parameters of the Beta distribution here as we'll seek to learn them during inference. The Beta distribution has two parameters: $\alpha$ and $\beta$, but I think it would be better to parameterise the likelihood in terms of it's mean $\mu = \Phi(f)$ where $\Phi$ is a Gaussian CDF, and dispersion $\nu$. This then makes the shape and scale parameters $\alpha=\nu\mu$ and $\beta=\nu(1-\mu)$. For more details on this, see Section 3.3 of [BOUNDED GAUSSIAN PROCESS REGRESSION](http://www2.imm.dtu.dk/pubdb/edoc/imm6637.pdf)

gpjax/likelihoods.py
```diff
@@ -199,6 +199,42 @@ def predict(self, dist: tfd.Distribution) -> tfd.Distribution:
...
+
+        Args:
+            alpha : alpha parameter, indicating mean number of successes
+            beta  : beta parameter, indicating mean number of failures
+
+        Returns
+        -------
+            tfd.Distribution: The likelihood function.
+        """
+        return tfd.Beta(alpha, beta)
```
thomaspinder, line 215: I think with the above change I've suggested, you'll need to call the inverse probit function in here, similar to the Bernoulli likelihood.

thomaspinder (changes requested):

gpjax/likelihoods.py
brandall25: The paper says *v* is an (inverse) dispersion parameter. I'm assuming dispersion refers to standard deviation, but what does it mean to take the inverse of it? Just the reciprocal?

gpjax/likelihoods.py
brandall25: Should the inverse probits be applied something like this:

`return tfd.Beta(concentration1=inv_probit(alpha), concentration1=inv_probit(beta))` ?

I'm not entirely sure what this does, is it just to normalize the alpha and beta parameters to [0, 1]?

gpjax/likelihoods.py
```diff
@@ -199,6 +199,44 @@ def predict(self, dist: tfd.Distribution) -> tfd.Distribution:
...
+
+        Args:
+            f (Float[Array, "..."]): Function values.
+
+        Returns
+        -------
+            tfd.Distribution: The likelihood function.
+        """
+
+        return tfd.Beta(inv_probit(f[0]), inv_probit(f[1]))
```
daniel-dodd, line 215: This is a step in the right direction, but note our variable `f` is vector of length $N$. In particular, this is vector of function values $\mathbf{f} = (f_1, \dotsc, f_N) \in \mathbb{R}^{N}$.

As such our Beta likelihood, $p(\mathbf{y}|\mathbf{f})$  models observations $\mathbf{y} = (y_1, \dotsc, y_N)$, that are drawn $y_i \sim \text{Beta}(\alpha_i, \beta_i)$ with $\mathbf{\alpha} = \nu \Phi (\mathbf{f})$, $\mathbf{\beta} = \nu (1 - \Phi (\mathbf{f}))$, $\mathbf{\alpha} = (\alpha_1, \dotsc, \alpha_N)$, $\mathbf{\beta} = (\beta_1, \dotsc, \beta_N)$. 

You need to pass these $\alpha$ and $\beta$ vectors to `tfd.Beta`.

For the dispersion parameter $\nu$, you'll need to mimic the `obs_noise` variable for our `Gaussian` likelihood.

gpjax/likelihoods.py
```diff
@@ -199,6 +199,44 @@ def predict(self, dist: tfd.Distribution) -> tfd.Distribution:
...
+        -------
+            tfd.Distribution: The pointwise predictive distribution.
+        """
+        mean = dist.mean().ravel()
+        stddev = dist.stddev().ravel()
+
+        alpha = mean / stddev
+        beta = (1.0 - mean) / stddev
+
+        return self.link_function([alpha, beta])
```
daniel-dodd, line 237: I don't believe the predictive marginal $p(y*) = \int p(y*| f*) p(f* | Data)$  can be done in closed form here the beta likelihood. (Or can it @thomaspinder?) If there is no analytical expression, you will need to either draw samples from the posterior distribution and use Monte, our use our quadrature integrator on the link functions's `log_prob` function $p(y_i|f_i)$.

daniel-dodd (changes requested):
Hi @brandall25, making nice progress, the thing that would be nice to focus on is getting that `link_function`. Leave that `predict` function for now. 

@thomaspinder does that `predict` function even have a closed form expression on this example? If not, would it be good for use to add a default on the base class that "integrates" this automatically, like the `expected_log_prob`?

gpjax/likelihoods.py
daniel-dodd: Hey @brandall25,

Thanks, good questions. I'll answer these in two separate comments.

> Something like this? I'm not sure f is needed in the link function.

Ah really we mean an inverse link function here, but we have rather relaxed this term.

We are looking for  for the map

$$f_i \mapsto p(y_i | g^{-1}(f_i))$$ 

where $g(\mu_i) = f_i$ and $\mu_i$ is the mean of the beta distribution, $g$ is the *link function* in the context of Generalised linear models, $y_i$ are the observations. So in particular, we are looking to reparameterise your code:

```python
def link_function(self, alpha: Float[Array, "..."], beta: Float[Array, "..."]) -> tfd.Distribution:
        r"""The probit link function of the Beta likelihood.
       ...

```

In terms of $\mathbf{f}$, as 

 ```python
def link_function(self, f: Float[Array, "N"]) -> tfd.Distribution:
        r"""The probit link function of the Beta likelihood.
       ...
```
And we need to use something to relate $\mathbf{f}$ to $\mathbf{\alpha}$ and $\mathbf{\beta}$ of the beta distribution. We suggested to do this with the $\Phi$ and dispersion parameter $\nu$.

Comment (daniel-dodd) [reply gpjax/likelihoods.py:215]: > Also, I'm not understanding what you mean by mimicking obs_noise for the dispersion parameter. You're saying v should be defined as a union like obs_noise`? Why is that?

Regarding the `obs_noise` term on the `Gaussian` likelihood - notice this is a special type of dataclass field, called a `param_field` as shown below: 

```python
@dataclass
class Gaussian(AbstractLikelihood):
    r"""Gaussian likelihood object."""

    obs_noise: Union[ScalarFloat, Float[Array, "#N"]] = param_field(
        jnp.array(1.0), bijector=tfb.Softplus()
    )
    ...
```

As such the `param_field` defines two things:

- (1) The default value the parameter should take, in this instance `jnp.array(1.0)` 
- (2) The domain  of the parameter through `bijector=tfb.Softplus()` - meaning this is a positive parameter. 

For the beta likelihood, we would need to have something that takes this same structure for our `dispersion` parameter $\nu$, as that too is a positive parameter, and you would need to do something like this.
```python
@dataclass
class Beta(AbstractLikelihood):
    r"""Beta likelihood object."""

    dispersion: Union[ScalarFloat, Float[Array, "#N"]] = param_field(
        jnp.array(1.0), bijector=tfb.Softplus()
    )
    ...
```

brandall25: So is `f` is the pdf of `dist`? 

It would be called from `predict` something like this `return self.link_function(dist.prob())`

And then `alpha` and `beta` are derived from `f` in `link_function`?

daniel-dodd: Hi @brandall25,

This is not quite the case no. I'll give a bit more detail on this below:

## --- Details ---

### `link_function`
- `f` parameterises the pdf of $y$ and this is what we want to return. This object is the conditional distribution $y | f$. 
- Specifically , we want to return the `tfd.Beta(alpha, beta)`, where `alpha = self.dispersion * inv_probit(f)` and `beta = self.dispersion * (1 - inv_probit(f))`. 

### `predict `
- The `dist` in  `predict` method will be an arbitrary Multivariate Gaussian $\mathbf{f}= (f_1, \dotsc , f_n) \sim N(\mu, \Sigma)$ where $\mu$ is a mean for $\mathbf{f}$ and $\Sigma$ is the covariance.  (This will actually be the predictive distribution for the function of interest `f`).

- To do the `predict`, you will need to draw samples from `dist`, $\mathbf{f}^{(j)} \sim  N(\mu, \Sigma)$, $j=1,..., M$, and return a Monte Carlo estimate -> this is a Mixture distribution $\mathbf{y}^* \sim \frac{1}{M} \sum_{j} p(\cdot| \mathbf{f}^{j}) $, which can be implemented with `tfd.Mixture` (see, https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/Mixture) . 
- As such,  `tfd.Mixture` has two arguments;  `components` that will be a list of `link_function`s evaluated at each $\mathbf{f}^{j}$ (so i.e., a list of the corresponding `tfd.Distribution` that each `link_function`($\mathbf{f}^{j}$) gives). And the other argument is a `cat` argument which should be uniform $\frac{1}{M}$ weights. 
- We actually need this `predict` method in general. So it would be nice to have this on the `AbstractLikelihood` class, to serve as the default option.
 
## --- Takeaway ---

- I strongly recommend focusing on the `link_function` for now, getting this coded in would be an excellent start. 
- Then, if you feel comfortable having a crack at the `predict` method, then please do so! Also I'm happy to help with it, just give me a shout.


Cheers,
Dan

thomaspinder:
Closing due to lack of activity

Issue #305, opened 2023-06-11
dev: Add regression smoke test


thomaspinder:
A [smoke test](https://en.wikipedia.org/wiki/Smoke_testing_(software)) for a regression example with the conjugate mll objective should be added. This can follow a similar workflow to the regression notebook where the final MLL, predictive mean, and predictive variance/standard deviation values are tested.

Issue #304, opened 2023-06-11
dev: Add stochastic VI smoke test


thomaspinder:
A [smoke test](https://en.wikipedia.org/wiki/Smoke_testing_(software)) for a stochastic VI example with the ELBO objective should be added. This can follow a similar workflow to the stochastic VI notebook where the final ELBO, predictive mean, and predictive variance/standard deviation values are tested.

Issue #303, opened 2023-06-11
dev: Add sparse GP smoke test


thomaspinder:
A [smoke test](https://en.wikipedia.org/wiki/Smoke_testing_(software)) for a sparse regression example with the collapsed ELBO should be added. This can follow a similar workflow to the sparse notebook where the final ELBO, predictive mean, and predictive variance/standard deviation values are tested.

PR #307, opened 2023-06-11
Smoke testing

thomaspinder:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [X] Tests
- [ ] Other

## Checklist

## Description

This PR introduces a smoke test for the regression workflow facilitated by GPJax. It asserts that the predictive mean, predictive standard deviation, MLL evaluate to ground truths.

Issue Number: #303 #304 #305


tests/smoke.py
```diff
@@ -0,0 +1,83 @@
...
+import jupytext
+import numpy as np
+
+
+@dataclass
+class Result:
+    path: str
+    true_mean: float
+    true_stddev: float
+    true_objective: float
```
daniel-dodd, line 13: - This is neat. I like it.

- Maybe though there is a better structure that allows for other things we can test. 
- To me the result is just a fixed dataclass of values: 
```python
@dataclass
 class Result:
     true_mean: float
     true_stddev: float
     true_objective: float
```

And we could have smoke test:
```python
@dataclass
 class SmokeTest:
     path: str
     
    def test(self,..., result: Result):
        ....
```
or 
```python
@dataclass
 class SmokeTest:
     path: str
     result: Result

    def test(self,..):
        ....
```
This would perhaps give a bit more flexibility.

thomaspinder: Sorry, you've lost me here. Why would this be better or more flexible than what we have?

st--:
very minor nit: if you actually check return values for correctness, it's not a smoke test, then it's a function integration test :wink: smoke test is just "run it and see that it doesn't blow up"...

tests/smoke.py
st--: how about rewriting it in a way that you can just write
```python
sparse = Result(
    path="docs/examples/collapsed_vi.py",
    history=1924.7634809,
    predictive_mean=-8.39869652,
    predictive_std=255.74838027,
)
sparse.test()

stochastic = Result(
    path="docs/examples/uncollapsed_vi.py",
    history=-985.71726453,
    meanf=-54.14787028,
    sigma=116.16651265,
)
stochastic.test()
```
?

st--: (and then it should just be an array instead of its sum for mean & std)

thomaspinder:
> very minor nit: if you actually check return values for correctness, it's not a smoke test, then it's a function integration test 😉 smoke test is just "run it and see that it doesn't blow up"...

Ahh, interesting. I always thought they were one of the same. Thanks!

daniel-dodd (approved):
Looks good to me @thomaspinder.

Issue #306, opened 2023-06-11
dev: Add benchmarking


thomaspinder:
Using something like [Airspeed velocity](https://asv.readthedocs.io/en/stable/index.html), the performance of GPJax should be monitored from one version to the next. Key objects to benchmark the performance of are the 

- [x] MLL
- [x] Log-posterior density
- [x] Collapsed ELBO
- [x] ELBO
- [x] Kernel gram matrix construction
- [x] Posterior predictive

PR #319, opened 2023-06-19
Benchmarking

thomaspinder:
## Type of changes

- [ ] Bug fix
- [X] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

This PR introduces benchmarking functionality to evaluate the runtime of core GPJax operations. The benchmarks are run every day at 1AM. The results are then formatted as html files and moved to the `benchmarks` branch which is then served on https://benchmarks.jaxgaussianprocesses.com/.

Issue Number: #306


.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,52 @@
...
+  workflow_dispatch:
+
+jobs:
+  run-benchmarks:
+    name: Run benchmarks
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: ["ubuntu-latest"]
+        python-version: ["3.8"]
```
Thomas-Christie, line 17: Will we be expanding this to also test against `macos-latest` and later Python versions, as in the other tests, in the future?

benchmarks/predictions.py
```diff
@@ -0,0 +1,86 @@
...
+        key = jr.PRNGKey(123)
+        self.X = jr.normal(key=key, shape=(100, n_dims))
+        self.y = jnp.sin(self.X[:, :1])
+        self.y = jnp.array(jnp.where(self.y > 0, 1, 0), dtype=jnp.float64)
+        self.data = gpx.Dataset(X=self.X, y=self.y)
+        kernel = gpx.kernels.RBF(active_dims=list(range(n_dims)))
+        meanf = gpx.mean_functions.Constant()
+        self.prior = gpx.Prior(kernel=kernel, mean_function=meanf)
+        self.likelihood = gpx.likelihoods.Bernoulli(num_datapoints=self.data.n)
+        self.posterior: ConjugatePosterior = self.prior * self.likelihood
```
Thomas-Christie, line 56: Minor question - won't this be a `NonConjugatePosterior`? Also not sure why there is a type annotation for this, but I could be missing something.

benchmarks/predictions.py
```diff
@@ -0,0 +1,86 @@
...
+        key = jr.PRNGKey(123)
+        self.X = jr.normal(key=key, shape=(100, n_dims))
+        f = lambda x: 2.0 * jnp.sin(3 * x) + 0.5 * x  # latent function
+        self.y = jnp.array(jr.poisson(key, jnp.exp(f(self.X))), dtype=jnp.float64)
+        self.data = gpx.Dataset(X=self.X, y=self.y)
+        kernel = gpx.kernels.RBF(active_dims=list(range(n_dims)))
+        meanf = gpx.mean_functions.Constant()
+        self.prior = gpx.Prior(kernel=kernel, mean_function=meanf)
+        self.likelihood = gpx.likelihoods.Bernoulli(num_datapoints=self.data.n)
+        self.posterior: ConjugatePosterior = self.prior * self.likelihood
```
Thomas-Christie, line 81: As above.

Thomas-Christie (approved):
Looks cool @thomaspinder ! I left a few very minor comments. I can't really assess how the visualisations will be because I can't see any graphs on [here](https://benchmarks.jaxgaussianprocesses.com/#/), but I assume we'll be able to see this once this gets merged. Also, I noticed that on https://github.com/JaxGaussianProcesses/GPJax/issues/306 the `ELBO` test isn't implemented yet. Are you planning on implementing this before merging?

benchmarks/kernels.py
```diff
@@ -0,0 +1,99 @@
...
+            key=key, minval=-3.0, maxval=3.0, shape=(n_datapoints, n_dims)
+        )
+
+
+class RBF(Kernels):
+    def setup(self, n_datapoints, n_dims):
+        super().setup(n_datapoints, n_dims)
+        self.kernel = kernels.RBF(active_dims=list(range(n_dims)))
+
+    def time_covfunc_call(self, n_datapoints, n_dims):
```
frazane, line 26: I suppose it doesn't work if we implement this on the parent class `Kernels` instead of on each subclass?

.github/workflows/benchmarks.yml
thomaspinder: I'm not sure the OS should be a variable in benchmarks. The key factor should be the delta version-on-version.

benchmarks/kernels.py
```diff
@@ -0,0 +1,99 @@
+from jax.config import config
```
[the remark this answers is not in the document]
thomaspinder: Thanks! I did not know about this!

benchmarks/predictions.py
thomaspinder: Ah, this is actually not an issue thanks to a PR @frazane merged in the time since this PR was created.

benchmarks/kernels.py
thomaspinder: Unfortunately not. That would be a cool feature.

PR #325, opened 2023-06-23
Nonstationary RFF

frazane:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Adding a new random features kernel based on https://www.sciencedirect.com/science/article/pii/S2211675317302890#b63. 

Issue Number: N/A

### DIFF
--- a/gpjax/__init__.py
+++ b/gpjax/__init__.py
@@ -38,6 +38,7 @@ from gpjax.kernels import (
     Matern12,
     Matern32,
     Matern52,
+    NonStationaryRFF,
     Periodic,
     Polynomial,
     PoweredExponential,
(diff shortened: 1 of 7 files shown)


thomaspinder:
Is this still a draft PR?

frazane:
> Is this still a draft PR?

Yes!

- I still have to add tests.
- Have to implement functional sampling.
- Maybe include the `NonStationaryRFF` to the `spatial.py` example?
- I wanted to discuss how the parameters are defined: as you can see, the current implementation doesn't have a `base_kernel` as in `RFF`, and parameters like `variance` and `lenghtscale` are directly defined on the `NonStationaryRFF` class. That's because I don't think this class would be used as an approximation of a base kernel, but rather as its own class of kernel, where one would learn frequencies (as they do in the paper).

thomaspinder:
> Maybe include the `NonStationaryRFF` to the spatial.py example?

Sounds good to me!

PR #328, opened 2023-06-27
Decouple kernel computation class initialisation from kernel

### LINKED ISSUE #293: dev: Decouple kernel computation class initialisation from kernel.
## Currently:
Currently the kernel computation is a dataclass that takes in the kernel as its argument.

```python
@dataclass
class AbstractKernelComputation:
    r"""Abstract class for kernel computations."""

    kernel: "gpjax.kernels.base.AbstractKernel"  # noqa: F821

```

This means when defining the `compute_engine` of a kernel, we cannot initialise the Computation before initialising the kernel, so have to pass an uninitialised class through ` compute_engine = DenseKernelComputation ` :

```python
@dataclass
class AbstractKernel(Module):
    r"""Base kernel class."""

    compute_engine: Type[AbstractKernelComputation] = static_field(
        DenseKernelComputation
    )
```
## Proposal:

- Remove the `kernel` argument on the `AbstractKernelComputation` so that the class can be initialised independently of the kernel.
- As such we would instead pass the `kernel` pytree, through the `gram`, `cross_covariance` methods of `AbstractKernelComputation`. 
- This would then allow us to pass `DenseKernelComputation()` instead of `DenseKernelComputation` to the `compute_engine`.

frazane:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [x] Internal refactoring

## Checklist

## Description

Closes #293


gpjax/kernels/computations/basis_functions.py
```diff
@@ -1,69 +1,76 @@
...
+        scaling_factor = kernel.base_kernel.lengthscale
         z = jnp.matmul(x, (frequencies / scaling_factor).T)
         z = jnp.concatenate([jnp.cos(z), jnp.sin(z)], axis=-1)
         return z
 
-    @property
-    def scaling(self):
-        return self.kernel.base_kernel.variance / self.kernel.num_basis_fns
+    def scaling(self, kernel):
+        return kernel.base_kernel.variance / kernel.num_basis_fns
```
daniel-dodd, line 76: A succinct docstring, and signature annotation would be good here.

gpjax/kernels/computations/eigen.py
```diff
@@ -33,14 +36,15 @@ class EigenKernelComputation(AbstractKernelComputation):
     """
 
     def cross_covariance(
-        self, x: Num[Array, "N D"], y: Num[Array, "M D"]
+        self, kernel: Kernel, x: Num[Array, "N D"], y: Num[Array, "M D"]
```
daniel-dodd, line 39: Noticed some of the signatures for `cross_covariance` and `gram` use `Num` instead of `Float`. Not sure if these were intended (or how they crept in if not)? But the were here prior to your PR, so can be addressed in a separate issue.

gpjax/kernels/computations/basis_functions.py
```diff
@@ -1,69 +1,76 @@
...
+from jaxtyping import (
+    Array,
+    Float,
+)
 
 from gpjax.kernels.computations.base import AbstractKernelComputation
 from gpjax.linops import DenseLinearOperator
-from gpjax.typing import Array
+
+Kernel = tp.TypeVar("Kernel", bound="gpjax.kernels.base.AbstractKernel")  # noqa: F821
```
daniel-dodd, line 13: Nice!

daniel-dodd (changes requested):
Hi @frazane, nice PR, exactly what we wanted to see for #328. 

Just had a single comment about a docstring / signature missing for a public method. And somethings up with the documentation test - will take a look at that shortly.

Aside, other-things, there are `Num` instead of `Float`'s in places that can be addressed in a separate PR. And I like that the computation uses the first slot as the input for the kernel PyTree, it would be good to mimic this convention for the likelihood integrator's, and to add your nice `TypeVar` annotation for the Likelihood in [integrators.py](https://github.com/JaxGaussianProcesses/GPJax/blob/main/gpjax/integrators.py) too. Will open an issues for these.

gpjax/kernels/computations/eigen.py
frazane: Yes I was also confused about that, and apparently it's because the `GraphKernel` implementation accepts integer types. I had all `Num`s changed to `Float` but beartype was complaining so I had to switch back...for now 😅

daniel-dodd (approved):

thomaspinder (approved):

thomaspinder:
LGTM. Happy for you to merge @frazane

PR #335, opened 2023-07-14
Add BO notebook

Thomas-Christie:
Added introduction to Bayesian optimisation notebook.

Also fixed some links which had broken after renaming the old kernels notebook in a previous PR.

## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [X] Other - New notebook

## Checklist

- [] I've added tests for new code - N/A
- [] I've added docstrings for the new code - N/A

## Description

Have added a new notebook which serves as an introduction to Bayesian optimisation. Also fixed a few links in the documentation which were broken when the original kernels notebook was renamed.

Issue Number: N/A


docs/examples/bayesian_optimisation.py
```diff
@@ -0,0 +1,737 @@
...
+# the objective function depending on the property being optimised, for instance within
+# drug-design this may be the efficacy of the drug. The function from molecules to
+# efficacy is unknown, but can be queried by synthesising a molecule and running an
+# experiment to measure its efficacy. This is clearly an expensive procedure!
+#
+# Within the domain of machine learning, the task of optimising neural network
+# architectures is another example of such a problem (commonly referred to as [Neural
+# Architecture Search (NAS)](https://en.wikipedia.org/wiki/Neural_architecture_search)).
+# Here, the domain is the space of possible neural network architectures, and the
+# objective function is a metric such as the accuracy of the trained model. Again, the
```
henrymoss, line 56: Lovely intro!

docs/examples/bayesian_optimisation.py
```diff
@@ -0,0 +1,737 @@
...
+# surrogate model, and uses the surrogate model's posterior distribution to quantify the
+# expected *utility*, $U$, of evaluating the black-box function at a given point. Simply
+# put, for each point in the black-box function's domain, $\mathbf{x} \in X$, the
+# acquisition function quantifies how useful it would be to evaluate the black-box
+# function at $\mathbf{x}$ in order to find the minimum of the black-box function, whilst
+# taking into consideration all the datapoints observed so far. Therefore, in order to
+# decide which point to query next we simply choose the point which maximises the
+# acquisition function, using an optimiser such as L-BFGS ([Liu and Nocedal,
+# 1989](https://link.springer.com/article/10.1007/BF01589116)) or Adam ([Kingma and Ba,
+# 2014](https://arxiv.org/abs/1412.6980)).
```
henrymoss, line 118: Using ADAM should not be encouraged!

docs/examples/bayesian_optimisation.py
```diff
@@ -0,0 +1,737 @@
...
+# provides an excellent overview.
+#
+# In this guide we will focus on *Thompson sampling*, a conceptually simple yet effective
+# method for characterising the utility of querying points in a black-box function's
+# domain, which will be useful in demonstrating the key aspects of BO.
+
+# %% [markdown]
+# ## Thompson Sampling
+#
+# Thompson sampling ([Thompson, 1933](https://shorturl.at/ejmCM)) is a simple method which
```
henrymoss, line 150: I have never referenced this. I definitely will in the future!

docs/examples/bayesian_optimisation.py
```diff
@@ -0,0 +1,737 @@
...
+
+
+# %% [markdown]
+# Next we'll define our GP model in the usual way, using a Matérn52 kernel, and fit the
+# kernel parameters by minimising the negative log-marginal likelihood. We'll wrap this in
+# a function as we'll be repeating this process at each iteration of the BO loop.
+
+
+# %%
+def generate_optimised_posterior(
```
henrymoss, line 204: I think I would prefer return_fitted_model or something else? We don't really generate posterior (also are all posteriors not optimised?).

docs/examples/bayesian_optimisation.py
```diff
@@ -0,0 +1,737 @@
...
+# Sampling Notebook](https://docs.jaxgaussianprocesses.com/examples/spatial/). This method
+# is used as it enables us to sample from the posterior in a manner which scales linearly
+# with the number of points sampled, $O(N)$, mitigating the cubic cost associated with
+# drawing exact samples from a GP posterior, $O(N^3)$. It also generates more accurate
+# samples than many other methods for drawing approximate samples from a GP posterior.
+#
+# Note that we also define a `utility_fn` which calls the approximate
+# sample but returns the value returned as a scalar. This is because the `sample_approx`
+# function returns an array of shape $[N, B]$, with $N$ being the number of points within
+# each sample and $B$ being the number of samples drawn. We'll only be drawing (and
```
henrymoss, line 251: this is definitely something we will hide away once we have some BO functionality

docs/examples/bayesian_optimisation.py
```diff
@@ -0,0 +1,737 @@
...
+        key,
+        shape=(num_initial_sample_points, lower_bound.shape[0]),
+        dtype=jnp.float64,
+        minval=lower_bound,
+        maxval=upper_bound,
+    )
+    initial_sample_y = sample(initial_sample_points)
+    best_x = jnp.array([initial_sample_points[jnp.argmin(initial_sample_y)]])
+
+    negative_utility_fn = lambda x: sample(x)[0][0]
```
henrymoss, line 296: maybe its worth stressing that we use negative_utility_fn because the optimizer is a minimiser?

docs/examples/bayesian_optimisation.py
```diff
@@ -0,0 +1,737 @@
...
+
+for i in range(bo_iters):
+    key, subkey = jr.split(key)
+
+    # Generate optimised posterior using previously observed data
+    mean = gpx.mean_functions.Zero()
+    kernel = gpx.kernels.Matern52()
+    prior = gpx.Prior(mean_function=mean, kernel=kernel)
+    opt_posterior = generate_optimised_posterior(D, prior, subkey)
+
```
henrymoss, line 405: maybe could just have a whole fit_model method that just takes in D?

docs/examples/bayesian_optimisation.py
```diff
@@ -0,0 +1,737 @@
...
+            num_samples=1, train_data=D, key=subkey, num_features=500
+        )
+        x_star = optimise_sample(
+            approx_sample,
+            subkey,
+            lower_bound,
+            upper_bound,
+            num_initial_sample_points=1000,
+        )
+
```
henrymoss, line 559: could even have a Thompson_Sample function that takes in a model and just return a chosen point?

docs/examples/bayesian_optimisation.py
```diff
@@ -0,0 +1,737 @@
...
+    D = D + gpx.Dataset(X=x_star, y=y_star)
+
+# %% [markdown]
+# Below we plot the best observed black-box function value against the number of times
+# the black-box function has been evaluated. Note that the first 5 samples are randomly
+# sampled to fit the initial GP model, and we denote the start of using BO to sample with
+# the dotted vertical line.
+#
+# We can see that the BO algorithm quickly converges to the global minimum of the
+# black-box function!
```
henrymoss, line 428: maybe just do 5 bo steps to speed up the notebook a bit

docs/examples/bayesian_optimisation.py
```diff
@@ -0,0 +1,737 @@
...
+)
+(
+    random_cumulative_min_mean,
+    random_cumulative_min_std,
+) = obtain_cumulative_minimum_statistics(random_experiment_results)
+
+# %% [markdown]
+# Now, when we plot the mean and standard deviation of the best observed value of the
+# black-box function at each iteration, we can see that BO outperforms random sampling,
+# consistently converging to the global minimum of the function.
```
henrymoss, line 629: we need this to be log(regret) so we can actually see a difference. I would be tempted to run for 15 steps to save a bit of time

henrymoss (approved):
This is fantastic! 
Just a couple small comments

docs/examples/bayesian_optimisation.py
Thomas-Christie: I think I'll rename to `return_optimised_posterior`. The reason I'm calling the posterior optimised is because of the kernel hyperparameter optimisation being performed. This is also in line with the other notebooks (e.g. https://docs.jaxgaussianprocesses.com/examples/barycentres/#learning-a-posterior-distribution)

docs/examples/bayesian_optimisation.py
Thomas-Christie: I had originally planned to do this, but in the second example end up having to use a slightly different prior (with two active dimensions and a slightly higher variance) so felt like being able to pass a prior was a bit easier in the end.

docs/examples/bayesian_optimisation.py
Thomas-Christie: Have changed to log(regret), and run for 16 steps to show a clear difference whilst saving time.

PR #338, opened 2023-07-28
[WIP] Add basic components for Bayesian optimisation

Thomas-Christie:
Add the core componenets for running BO in GPJax.

## Type of changes

- [ ] Bug fix
- [X] New feature
- [ ] Tests
- [ ] Other

## Checklist

- [ ] I've added tests for new code (WIP - will be done in subsequent commit)

## Description

Added some code for a BO framework within GPJax. This can be explored in the `bo_demo.ipynb` notebook (will be removed before merging). 

Key TODOs:
- Decide on how we'll deal with the `AbstractLikelihood` object requiring a number of datapoints to be specified, as this will change throughout the BO loop and ideally we don't want to be having to reconstruct the posterior object on each iteration of the loop.
- Add tests.
- Figure out how to deal with type annotation for functions provided in `post_ask` and `post_tell` in `BayesianOptimiser`. This seems to be a bit of a chicken and egg problem, with `BayesianOptimiser` being passed to these methods as arguments (but if we annotate with this type then we'll be referring to this type from within the definition of the `BayesianOptimiser`class). 
- Think about if we'll also want some kind of attribute in the `BayesianOptimiser` class which can be used to store more state (if the user wants to deal with more state in their `post_ask` and `post_tell` methods).

Issue Number: N/A


gpjax/bayes_opt/acquisition_optimiser.py
```diff
@@ -0,0 +1,111 @@
...
+    KeyArray,
+)
+
+
+@dataclass
+class AbstractAcquisitionOptimiser(ABC):
+    """Abstract base class for acquisition function optimisers."""
+
+    @abstractmethod
+    def optimise(
```
thomaspinder, line 42: Design choice - do we want `optimise` or `minimise`? The latter is nice as it connect with the API in other optimisation libraries like Scipy. Further, we may wish to define `optimize` or `minimize` here to simply call the respective method. It makes the package more accessible to people using American-English.

gpjax/bayes_opt/acquisition_optimiser.py
```diff
@@ -0,0 +1,111 @@
...
+
+        best_initial_sample_value_idx = jnp.argmax(
+            initial_sample_values, axis=0, keepdims=True
+        )
+        best_initial_sample_point = jnp.take_along_axis(
+            initial_sample_points, best_initial_sample_value_idx, axis=0
+        )  # [1, D]
+        print(
+            f"Best Initial Sample Point Shape: {best_initial_sample_point.shape} Should be [1, D]"
+        )
```
thomaspinder, line 97: Do you want these print statements in production code?

gpjax/bayes_opt/acquisition_optimiser.py
```diff
@@ -0,0 +1,111 @@
...
+        print(
+            f"Initial Sample Values Shape: {initial_sample_values.shape} Should be [N, 1]"
+        )
+
+        best_initial_sample_value_idx = jnp.argmax(
+            initial_sample_values, axis=0, keepdims=True
+        )
+        best_initial_sample_point = jnp.take_along_axis(
+            initial_sample_points, best_initial_sample_value_idx, axis=0
+        )  # [1, D]
```
thomaspinder, line 94: Could this be abstracted in another method named something like `get_max_sample_values()` or something? It may be a personal thing, but I always find `argmax`, `take_along_axis`.etc operations very hard to get the intuition of in isolation.

gpjax/bayes_opt/bayesian_optimiser.py
```diff
@@ -0,0 +1,117 @@
...
+    This can be run as a typical ask-tell loop, or the `run` method can be used to run
+    the optimisation for a fixed number of steps. Moreover, the `run` method executes
+    the functions in `post_ask` and `post_tell` after each ask and tell step
+    respectively. This enables the user to add custom functionality, such as the ability
+    to plot values of interest during the optimisation process.
+    """
+
+    search_space: AbstractSearchSpace
+    posteriors: Mapping[str, AbstractPosterior]
+    datasets: Mapping[str, Dataset]
```
thomaspinder, line 46: Why plural? We have just one dataset no?

thomaspinder (commented):
Nice work @Thomas-Christie . This is starting to come together really well. I'll be honest, there's a lot here and it's possible that I've missed some details so it'd be good to have a short walkthrough once you're back.

bo_demo.py
```diff
@@ -0,0 +1,146 @@
...
+    plt.show()
+
+
+if __name__ == "__main__":
+    key = jr.PRNGKey(42)
+    function_evaluator = build_function_evaluator({OBJECTIVE: forrester})
+    lower_bounds = jnp.array([0.0])
+    upper_bounds = jnp.array([1.0])
+    print(f"Lower Bounds Shape: {lower_bounds.shape}")
+    search_space = BoxSearchSpace(lower_bounds=lower_bounds, upper_bounds=upper_bounds)
```
henrymoss, line 110: I wonder, if the search space bounds could be in the dataset?

bo_demo.py
```diff
@@ -0,0 +1,146 @@
...
+    initial_x = search_space.sample_halton(5, key)
```
henrymoss, line 111: its looking a bit too much like Trieste here!

Thomas-Christie:
Closing as this work was implemented in https://github.com/JaxGaussianProcesses/GPJax/pull/386.

PR #340, opened 2023-07-31
Support for missing values using masked dataset

frazane:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

This PR introduces support for missing data points (missing `y`s) in a way that is compatible with JAX jit-compilation (by keeping shapes static). This could be particularly useful in meta-learning settings (where each task has a variable number of available training points), or in multi-output GPs where only some of the outputs are missing (@ingmarschuster).

- new attribute `mask` added to the `Dataset`object
- added code to computations of `log_prob` and to posterior predictions, such that masked training inputs have no influence on the results

For now I only tested this on a simple exact GP with gaussian likelihood.

### DIFF
--- a/gpjax/dataset.py
+++ b/gpjax/dataset.py
@@ -45,7 +45,7 @@ class Dataset(Pytree):
         _check_shape(self.X, self.y)
         if self.y is not None and self.mask is None:
             if jnp.any(mask := jnp.isnan(self.y)):
-                self.mask = ~mask
+                self.mask = mask
 
     def __repr__(self) -> str:
         r"""Returns a string representation of the dataset."""
--- a/gpjax/dataset.py
+++ b/gpjax/dataset.py
@@ -26,26 +26,42 @@ from simple_pytree import Pytree
 from gpjax.typing import Array
 
 
+class _Missing:
+    """Sentinel class for not-yet-computed mask"""
+
+
+
 @dataclass
 class Dataset(Pytree):
     r"""Base class for datasets.
 
     Attributes
     ----------
-        X (Optional[Num[Array, "N D"]]): Input data.
-        y (Optional[Num[Array, "N Q"]]): Output data.
+        X: Num[Array, "N D"], optional:
+            Input data.
+        y: Num[Array, "N Q"], optional:
+            Output data.
+        mask: Bool[Array, "N Q"] or None, optional
+            Mask for the output data. By default, a mask will be computed based on the output data.
+            User can optionally specify a pre-computed mask, or pass `None` which means no mask
+            will be computed.
     """
 
     X: Optional[Num[Array, "N D"]] = None
     y: Optional[Num[Array, "N Q"]] = None
-    mask: Optional[Bool[Array, "N Q"] | bool] = None
+    mask: Bool[Array, "N Q"] | None = _Missing()
 
     def __post_init__(self) -> None:
         r"""Checks that the shapes of $`X`$ and $`y`$ are compatible."""
         _check_shape(self.X, self.y)
-        if self.y is not None and self.mask is None:
-            if jnp.any(mask := jnp.isnan(self.y)):
-                self.mask = mask
+        if isinstance(self.mask, _Missing):
+            if self.y is not None:
+                if jnp.any(mask := jnp.isnan(self.y)):
+                    self.mask = mask
+                else:
+                    self.mask = None
+            else:
+                self.mask = None
 
     def __repr__(self) -> str:
         r"""Returns a string representation of the dataset."""
@@ -76,8 +92,8 @@ class Dataset(Pytree):
             y = jnp.concatenate((self.y, other.y))
 
         if (sm := self.mask is not None) | (om := other.mask is not None):
-            sm = self.mask if sm else jnp.ones(self.y.shape, dtype=bool)
-            om = other.mask if om else jnp.ones(other.y.shape, dtype=bool)
+            sm = self.mask if sm else jnp.zeros(self.y.shape, dtype=bool)
+            om = other.mask if om else jnp.zeros(other.y.shape, dtype=bool)
             mask = jnp.concatenate((sm, om))
 
         return Dataset(X=X, y=y, mask=mask)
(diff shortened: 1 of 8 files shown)


gpjax/dataset.py
```diff
@@ -15,30 +15,55 @@
...
     X: Optional[Num[Array, "N D"]] = None
     y: Optional[Num[Array, "N Q"]] = None
+    mask: Optional[Union[Bool[Array, "N Q"], None]] = _Missing()
 
     def __post_init__(self) -> None:
         r"""Checks that the shapes of $`X`$ and $`y`$ are compatible."""
         _check_shape(self.X, self.y)
+        if isinstance(self.mask, _Missing):
+            if self.y is not None:
+                if jnp.any(mask := jnp.isnan(self.y)):
```
ingmarschuster, line 61: I find
```
mask = jnp.isnan(self.y)
if jnp.any(mask)
```
easier to read than the := operator. this hold for later as well but is mostly a question of style.

gpjax/gaussian_distribution.py
```diff
@@ -149,7 +152,9 @@ def entropy(self) -> ScalarFloat:
             self.event_shape[0] * (1.0 + jnp.log(2.0 * jnp.pi)) + self.scale.log_det()
         )
 
-    def log_prob(self, y: Float[Array, " N"]) -> ScalarFloat:
+    def log_prob(
```
ingmarschuster, line 155: It would be great to add a test just for the masked case in `test_gaussian_distribution.py`. This way if something is off in this core functionality, one instantly knows.

ingmarschuster (changes requested):
Mostly needed is a test for masked log_prob. The := operator is just a question of style

gpjax/dataset.py
frazane: The walrus operator is still a controversial feature after all 😅 I agree it's not really needed in this case but later it makes for a cleaner code imo

gpjax/gaussian_distribution.py
frazane: Yes, I have to add this and other tests.

gpjax/dataset.py
thomaspinder: I think I agree with Ingmar here.

gpjax/gaussian_distribution.py
thomaspinder: +1

gpjax/dataset.py
```diff
@@ -15,30 +15,55 @@
...
+    Bool,
+    Num,
+)
 from simple_pytree import Pytree
 
 from gpjax.typing import Array
 
 
+class _Missing:
+    """Sentinel class for not-yet-computed mask"""
```
thomaspinder, line 33: Could you expand a little on this description please? To me, it doesn't fully explain the class' purpose.

Also, what is a sentinel class?

thomaspinder: On reflection, why do we even need this class?

gpjax/dataset.py
```diff
@@ -60,14 +85,20 @@ def __add__(self, other: "Dataset") -> "Dataset":
...
+        mask = None
 
         if self.X is not None and other.X is not None:
             X = jnp.concatenate((self.X, other.X))
 
         if self.y is not None and other.y is not None:
             y = jnp.concatenate((self.y, other.y))
 
-        return Dataset(X=X, y=y)
+        if (sm := self.mask is not None) | (om := other.mask is not None):
```
thomaspinder, line 96: Stylistic, but elsewhere in the package we use `or` not `|`. Could we be consistent here please unless there's a performance benefit? Above discussion on `:=` holds true here too.

thomaspinder (changes requested):

gpjax/dataset.py
frazane: The `Dataset` class holds either a boolean array or `None`, however at initialization `jnp.isnan` will be computed every time to ensure that there are no missing values in the data. This can be a lot of overhead especially when constructing batches (see `get_batch` in `fit.py`): if the original Dataset has `mask=None` there is no reason to do the computation for the subset. So the logic here is:
- if nothing is specified the default is this sentinel class, we compute the mask if needed
- if `None` is passed explicitly, we skip the computation and the mask is just `None`

In short, we need this class to provide a clear differentiation between an intentionally provided value of `None` and a default "unset" value for the mask attribute. 

In these situations the "sentinel object/value" pattern is useful, see https://python-patterns.guide/python/sentinel-object/

gpjax/dataset.py
frazane: "or" and "|" are not the same. If there are two conditions:
`cond1 or cond2`: cond2 does NOT evaluate if cond1 is already True
`cond1 | cond2`: both conditions are evaluated, and then True is returned is one of them is True (which is what we want here)

gpjax/dataset.py
ingmarschuster: This looks better to me than the solution I've always used (introduce an additional boolean variable `use_mask`). Maybe yet another would be to have a "Dataset" and a "MaskedDataset" class. I would be ok with the sentinel.

ingmarschuster (approved):
Sentinel would be fine to me. But see the comment about this above for other solution options I see

gpjax/dataset.py
ingmarschuster: another solution would be to use string constants, which is what some python libs do and you don't have to introduce a new class for it.

```
@dataclass
class Dataset:
    mask: Optional[Union[Bool[Array, "N Q"], str]] = "No Mask"
```

one hint: `Optional[...]` is a shorthand notation for `Union[..., None]` so your original type hint (` mask: Optional[Union[Bool[Array, "N Q"], None]] = _Missing()`) is doubling things.

ingmarschuster:
I changed the sentinel class `_Missing` to a string literal alternative. I believe this serves the same purpose and might be more GPJax-style simple.
Accidentally pushed to the upstream branch. @frazane and @thomaspinder, what do you think? Shall we merge this version?

Issue #342, opened 2023-08-04
dev: Add flag in Dataset when data is not float64


thomaspinder:
Add a warning in the Dataset class when the supplied arrays are not float64. Acknowledge that using float32 may be fine but it could cause unstable matrix operations downstream.

PR #347, opened 2023-08-06
Precision check feature

trsav:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [x] Tests
- [ ] Other

## Checklist

## Description

Have added a warning using the ```warnings``` package that informs when X or y isn't of dtype float64 (if the warnings package isn't appropriate then please let me know, it's a new import within ```dataset.py```). This is called immediately after checking the dataset shape:

```
    def __post_init__(self) -> None:
        r"""Checks that the shapes of $`X`$ and $`y`$ are compatible,
        and provides warnings regarding the precision of $`X`$ and $`y`$."""
        _check_shape(self.X, self.y)
       -->>  _check_precision(self.X, self.y)
```

I've added a test for the warning generation, but to do so I had to enable the higher precision within ```test_dataset.py``` (```config.update("jax_enable_x64", True)```) .  I hope this solution is alright. Thanks!

Issue Number: #342

### DIFF
--- a/gpjax/dataset.py
+++ b/gpjax/dataset.py
@@ -14,6 +14,7 @@
 # ==============================================================================
 
 from dataclasses import dataclass
+import warnings
 
 from beartype.typing import Optional
 import jax.numpy as jnp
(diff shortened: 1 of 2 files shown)


thomaspinder (approved):

thomaspinder:
Wonderful. Thanks for such a clean PR @trsav!

PR #351, opened 2023-08-08
[WIP] Remove `num_datapoints` attribute from `Likelihood` class

Thomas-Christie:
It would be highly beneficial not to have a `num_datapoints` attribute associated with likelihood functions, particularly for BO functionality. This is largely used in two places:

1. `NonConjugatePosterior`  - here it is used to initialise the latent function. I propose we can remove this by adding a `get_latent()` function and making the `latent` attribute private. Now, if `get_latent()`  is called, and the `latent` function hasn't been initialised, it will be initialised with the correct number of datapoints. As far as I'm aware, the number of datapoints should be available whenever this is called.

2. `ELBO` objective - here it is used to rescale the objective if batching is used. Instead, I propose that we add a `full_dataset_size` attribute to the `ELBO` class which is supplied by the user upon initialisation (at which point the size of the full dataset should be known).

## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [X] Other

## Checklist

- [ ] I've added tests for new code - Will be added in subsequent commit.

## Description

It would be highly beneficial not to have a `num_datapoints` attribute associated with likelihood functions, particularly for BO functionality. This is largely used in two places:

1. `NonConjugatePosterior`  - here it is used to initialise the latent function. I propose we can remove this by adding a `get_latent()` function and making the `latent` attribute private. Now, if `get_latent()`  is called, and the `latent` function hasn't been initialised, it will be initialised with the correct number of datapoints. As far as I'm aware, the number of datapoints should be available whenever this is called.

2. `ELBO` objective - here it is used to rescale the objective if batching is used. Instead, I propose that we add a `full_dataset_size` attribute to the `ELBO` class which is supplied by the user upon initialisation (at which point the size of the full dataset should be known).

Issue Number: N/A

### DIFF
--- a/gpjax/gps.py
+++ b/gpjax/gps.py
@@ -611,12 +611,17 @@ class NonConjugatePosterior(AbstractPosterior):
     from, or optimise an approximation to, the posterior distribution.
     """
 
-    latent: Float[Array, "N 1"] = param_field(None)
+    _latent: Float[Array, "N 1"] = param_field(None)
     key: KeyArray = static_field(PRNGKey(42))
 
-    def __post_init__(self):
-        if self.latent is None:
-            self.latent = normal(self.key, shape=(self.likelihood.num_datapoints, 1))
+    def get_latent(self, num_datapoints: int) -> Float[Array, "N 1"]:
+        if self._latent is None:
+            self._latent = normal(self.key, shape=(num_datapoints, 1))
+        elif self._latent.shape[0] != num_datapoints:
+            raise ValueError(
+                "The number of datapoints supplied is not equal to the number of datapoints used in the existing latent function."
+            )
+        return self._latent
 
     def predict(
         self, test_inputs: Num[Array, "N D"], train_data: Dataset
@@ -664,7 +669,7 @@ class NonConjugatePosterior(AbstractPosterior):
         Lx_inv_Kxt = Lx.solve(Ktx.T)
 
         # Whitened function values, wx, corresponding to the inputs, x
-        wx = self.latent
+        wx = self.get_latent(num_datapoints=n)
 
         # μt + Ktx Lx⁻¹ wx
         mean = mean_t + jnp.matmul(Lx_inv_Kxt.T, wx)
--- a/gpjax/kernels/__init__.py
+++ b/gpjax/kernels/__init__.py
@@ -27,7 +27,10 @@ from gpjax.kernels.computations import (
     DiagonalKernelComputation,
     EigenKernelComputation,
 )
-from gpjax.kernels.non_euclidean import GraphKernel, CatKernel
+from gpjax.kernels.non_euclidean import (
+    CatKernel,
+    GraphKernel,
+)
 from gpjax.kernels.nonstationary import (
     ArcCosine,
     Linear,
--- a/gpjax/kernels/non_euclidean/__init__.py
+++ b/gpjax/kernels/non_euclidean/__init__.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 # ==============================================================================
 
-from gpjax.kernels.non_euclidean.graph import GraphKernel
 from gpjax.kernels.non_euclidean.categorical import CatKernel
+from gpjax.kernels.non_euclidean.graph import GraphKernel
 
 __all__ = ["GraphKernel", "CatKernel"]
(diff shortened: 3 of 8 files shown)


gpjax/gps.py
```diff
@@ -664,7 +669,7 @@ def predict(
         Lx_inv_Kxt = Lx.solve(Ktx.T)
 
         # Whitened function values, wx, corresponding to the inputs, x
-        wx = self.latent
+        wx = self.get_latent(num_datapoints=n)
```
henrymoss, line 672: Surely self._latent is already built (i.e not None) at this point.  passing n doesnt do anything?  Same below as well

henrymoss: perhaps n should be optional ?

Thomas-Christie: There is no longer a guarantee that `self._latent` will be built at this point as there is no longer a `__post_init__` method (which relied on the `num_datapoints` field of the likelihood). I think that making `n` not optional could also be beneficial for ensuring that the number of datapoints used in the dataset and latent function are also consistent?

gpjax/gps.py
```diff
@@ -611,12 +611,17 @@ class NonConjugatePosterior(AbstractPosterior):
...
-        if self.latent is None:
-            self.latent = normal(self.key, shape=(self.likelihood.num_datapoints, 1))
+    def get_latent(self, num_datapoints: int) -> Float[Array, "N 1"]:
+        if self._latent is None:
+            self._latent = normal(self.key, shape=(num_datapoints, 1))
+        elif self._latent.shape[0] != num_datapoints:
+            raise ValueError(
+                "The number of datapoints supplied is not equal to the number of datapoints used in the existing latent function."
+            )
+        return self._latent
```
daniel-dodd, line 624: Careful. This look like a stateful method to me - which means not compatible with JAX tracers.

Also, I am surprised that `self._latent = normal...` assignment does not throw an error, as the `Module` should be immutable?

Thomas-Christie: Thanks for the review @daniel-dodd. For your latter point - it may well throw an error, I haven't actually tested it yet, was just looking for some feedback on the approach before refactoring all tests etc. 

For the former point, I see this could be a problem, and the side-effects didn't feel ideal. Do you think a better approach would be to simply do something like `return normal(self.key, shape=(num_datapoints, 1))` inside `get_latent()` function, and stop having a `latent` attribute? However, I guess this might be not ideal, as if the number of datapoints changes, the latent function values will change (I'm not sure if this lack of "consistency" is an issue though as I'm not too familiar with this NonConjugatePosterior stuff).

thomaspinder: I agree with @daniel-dodd that we likely have impure functions here which is going to break tracers. Further, the issue with not having the latent function's values stored as a field of the model is likely to make it tricky, if not impossible, to differentiate against the latent function's values.

thomaspinder: Maybe there's something more clever than I'm capable of conjuring up though that would solve this.

gpjax/gps.py
thomaspinder: Agreed with @Thomas-Christie here that under this approach, we'd need such an argument. The real crux of this is the earlier thread with Dan and Thomas.

gpjax/gps.py
daniel-dodd: Yeah, this is a tricky one. 

From my experience using GPJax for experimental design, specifying data point numbers can be frustrating. I typically worked with sparse GPs to independently set the model dimensions (via inducing points) of the datapoints dimension and could just pass e.g., approaches like `likelihood(num_datapoints=123)`. While not universally perfect, it resolved issues and aligned with scalability needs.

To initialise the NonConjugatePosterior model state, knowing the abstract data shape is crucial. This determines initial values for the latent variable, refining during training. This process parallels flax's `.init(key, dummy_x)` concept. The code structure currently demands awareness of the "`dummy_x`" shape's placement. Alternately, in GPFlow, posterior models are defined by upfront data provision: `model = gpflow.models.GPR(data, ...)` or `model = gpflow.models.GPMC(data,...)`. Is this data-forward approach more convenient than passing data to an objective function and abstract form (in terms of the datapoint number specification) to the likelihood?"

Thomas-Christie:
Having discussed offline, we decided to leave `num_datapoints` as an attribute of the `Likelihood` class, and so this PR will be closed.

PR #357, opened 2023-08-10
Add changepoint kernel

jakeyeung:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Add change point kernel for two RBFs. 

Issue Number: N/A
Related to Q&A in Discussion #337


gpjax/kernels/stationary/changepoint.py
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env python
+"""
+AUTHOR: Jake Yeung (jakeyeung@gmail.com)
+CREATED ON:  2023-07-28
+LAST CHANGE: see git log
+LICENSE:     Apache License
+"""
+
```
thomaspinder, line 8: Could you change this to the standard GPJax boilerplate please?

gpjax/kernels/stationary/changepoint.py
```diff
@@ -0,0 +1,117 @@
...
+import jax.numpy as jnp
+from jaxtyping import Float
+import tensorflow_probability.substrates.jax.bijectors as tfb
+
+import gpjax as gpx
+from gpjax.base.param import param_field
+from gpjax.typing import (
+    Array,
+    ScalarFloat,
+)
```
thomaspinder, line 22: These imports don't look sorted? Did the precommit successfully run?

gpjax/kernels/stationary/changepoint.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+    lengthscale1: Union[ScalarFloat, Float[Array, " D"]] = param_field(
+        jnp.array(1.0), bijector=tfb.Softplus(), trainable=True
+    )
+    variance2: ScalarFloat = param_field(
+        jnp.array(1.0), bijector=tfb.Softplus(), trainable=True
+    )
+    lengthscale2: Union[ScalarFloat, Float[Array, " D"]] = param_field(
+        jnp.array(1.0), bijector=tfb.Softplus(), trainable=True
+    )
```
thomaspinder, line 40: Could the constructor here simply accept two kernels as input? The parameters would then be lifted out of the substituent kernels.

gpjax/kernels/stationary/changepoint.py
```diff
@@ -0,0 +1,117 @@
...
+    tswitch: float = param_field(
```
thomaspinder, line 41: It would be good to acknowledge what this is in the docstring.

gpjax/kernels/stationary/changepoint.py
```diff
@@ -0,0 +1,117 @@
...
+from gpjax.base.param import param_field
+from gpjax.typing import (
+    Array,
+    ScalarFloat,
+)
+
+
+@dataclass
+class ChangePointRBF(gpx.kernels.AbstractKernel):
+    r"""The change point kernel for two RBFs."""
```
thomaspinder, line 27: Having some further details here would be helpful. At the minimum, explain what the kernel does and define the arguments required for initialisation.

gpjax/kernels/stationary/changepoint.py
```diff
@@ -0,0 +1,117 @@
...
+        ```
+
+        Args:
+            x (Float[Array, " N"]): The left hand argument of the kernel function's call.
+            y (Float[Array, " N"]): The right hand argument of the kernel function's call.
+            self.tswitch: The point at which to change to a different kernel.
+            self.variance1: Variance parameter for first RBF (if x, y both less than tswitch)
+            self.lengthscale1: Lengthscale parameter for first RBF (if x, y both less than tswitch)
+            self.variance2: Variance parameter for second RBF (if x, y both greater than or equal to tswitch)
+            self.lengthscale2: Lengthscale parameter for second RBF (if x, y both greater than or equal to tswitch)
```
thomaspinder, line 57: These lines should be move to the objects initialisation.

gpjax/kernels/stationary/changepoint.py
```diff
@@ -0,0 +1,117 @@
...
+
+        Returns:
+            ScalarFloat: The value of $`k(x, y)`. Three possible scenarios:
+            (1) RBF with parameters variance1, lengthscale1 if both x and y are less than tswitch
+            (2) RBF with parameters variance2, lengthscale2 is both x and y are greater than or equal to tswitch
+            (3) 0 otherwise$.
```
thomaspinder, line 63: I would add this information to the docstring's descriptor.

gpjax/kernels/stationary/changepoint.py
```diff
@@ -0,0 +1,117 @@
...
+
+        def k12(x1, x2, v1=1, l1=0.1, v2=2, l2=0.2):
+            r"""Return 0 covariance"""
+            out = jnp.float64(0)
+            return out.squeeze()
+
+        def k21(x1, x2, v1=1, l1=0.1, v2=2, l2=0.2):
+            r"""Return 0 covariance"""
+            out = jnp.float64(0)
+            return out.squeeze()
```
thomaspinder, line 84: If you take my suggestion of supplying the individual kernels to the class, then these lines can be removed.

gpjax/kernels/stationary/changepoint.py
```diff
@@ -0,0 +1,117 @@
...
+
+        def GetFunctionIndex(x, y, tswitch=1):
```
thomaspinder, line 86: nit: this should be lower case with words separated by `_`

thomaspinder (changes requested):
Nice work @jakeyeung ! I've left some comments that we should resolve before merging. It will also need some unit tests before we're able to merge.

gpjax/kernels/stationary/changepoint.py
jakeyeung: I checked with `isort` and it looks like this is how `isort` likes to organize the imports. 

Quickly checking other kernels like RBF https://github.com/JaxGaussianProcesses/GPJax/blob/main/gpjax/kernels/stationary/rbf.py suggests these imports are consistent.

gpjax/kernels/stationary/changepoint.py
jakeyeung: The initialization is now a list of two kernels, which hopefully simplifies the details here.

gpjax/kernels/stationary/changepoint.py
jakeyeung: I adapted the CombinationKernel and should now accept a list of two kernels as input.

gpjax/kernels/stationary/changepoint.py
jakeyeung: I added what tswitch does in the initialization of the `SwitchKernel`

gpjax/kernels/stationary/changepoint.py
jakeyeung: It's now moved, and simplified to just have a list of kernels and tswitch.

gpjax/kernels/stationary/changepoint.py
jakeyeung: Now removed.

gpjax/kernels/stationary/changepoint.py
jakeyeung: Now changed.

gpjax/kernels/stationary/changepoint.py
jakeyeung: Described these three scenarios under the initialization of the class, under `self.tswitch`

thomaspinder:
Nice work @jakeyeung - this PR is looking good. If we can add some unit tests to this, then I think we'll be good to merge.

jakeyeung:
> Nice work @jakeyeung - this PR is looking good. If we can add some unit tests to this, then I think we'll be good to merge.

Hi @thomaspinder, glad to know it's progressing. Any ideas of some unit tests that would be good, both ones standard for all kernels as well as ones specific to the changepoint?

jakeyeung:
Hi @thomaspinder, sorry for the delay. I have now added a `test_changepoint_kernel()` function under `test_stationary.py`. Hope everything is alright now to merge.

jakeyeung:
> Nice work @jakeyeung! Just to check: have you rebased from the main branch - the change log seems huge here? Would just make it easier to see what you have added. :)

Hi Daniel, yeah I think what happened was that `git fetch upstream` and `git rebase upstream/main` updated a bunch of files, and then after `poetry run pre-commit run --all-files` changed a lot of files that weren't added by me. 

I will fix this tomorrow and just commit the files that I changed.

gpjax/kernels/stationary/changepoint.py
```diff
@@ -0,0 +1,115 @@
...
+        for example: if x and y are both less than tswitch, then you would use kernels[0]
+                     if x and y are both greater than or equal to tswitch, then you would use kernels[1]
+                     otherwise return cross-covariance of 0
+    """
+
+    kernels: List[AbstractKernel] = None
+    operator: Callable = static_field(None)
+    tswitch: ScalarFloat = param_field(
+        jnp.array(1.0), bijector=tfb.Identity(), trainable=False
+    )
```
daniel-dodd, line 53: Is the changepoint location itself not learnt here?

Also what happens if we have multiple changepoints (does this kernel work recursively by passing a ChangePoint kernel to `kernels`?)

gpjax/kernels/stationary/changepoint.py
```diff
@@ -0,0 +1,115 @@
...
+from gpjax.kernels.base import AbstractKernel
+from gpjax.typing import (
+    Array,
+    ScalarFloat,
+)
+
+
+@dataclass
+class ChangePoint(AbstractKernel):
+    r"""A change point kernel
```
daniel-dodd, line 41: Is there a citation for your implementation of the kernel? We could add this to our `citation.py`. Also probably good to give a paper in the docstring. :)

gpjax/kernels/stationary/changepoint.py
```diff
@@ -0,0 +1,115 @@
...
+            else:
+                kernels_list.append(kernel)
+
+        self.kernels = kernels_list
+
+    def __call__(
+        self,
+        x: Float[Array, " D"],
+        y: Float[Array, " D"],
+    ) -> ScalarFloat:
```
daniel-dodd, line 74: Maybe I'd be clear the intended dimensions is one here? (correct me if I'm wrong). i.e., `Float[Array, " 1"]`  instead of `Float[Array, " D"]` .

gpjax/kernels/stationary/changepoint.py
```diff
@@ -0,0 +1,115 @@
...
+        Args:
+            x (Float[Array, " D"]): The left hand input of the kernel function.
+            y (Float[Array, " D"]): The right hand input of the kernel function.
+
+        Returns
+        -------
+            ScalarFloat: The evaluated kernel function at the supplied inputs.
+        """
+
+        def get_function_index(x, y, tswitch=1):
```
daniel-dodd, line 86: I would probably drop the default argument here.

```suggestion

        def get_function_index(x, y, tswitch):
```

gpjax/kernels/stationary/changepoint.py
```diff
@@ -0,0 +1,115 @@
...
+            cond2 = jnp.less(y, tswitch)
+            indx = jnp.where(jnp.logical_and(cond1, cond2), 0, indx)
+            indx = jnp.where(jnp.logical_and(jnp.invert(cond1), cond2), 1, indx)
+            indx = jnp.where(jnp.logical_and(cond1, jnp.invert(cond2)), 2, indx)
+            return indx.squeeze().astype("uint8")
+
+        def k_zero(x, y):
+            r"""Return 0 covariance"""
+            out = jnp.float64(0)
+            return out.squeeze()
```
daniel-dodd, line 108: This is nice. But the only concern I have is with `jnp.float64` potentially clashing with JAX's default `float32`. Could this not work with jnp.array(0.0) (or even 0.0)? Or does this break the `jax.lax.switch` or something? :)

daniel-dodd: If problems persist. You could always use a `Constant(0.0)` kernel here.

gpjax/kernels/stationary/changepoint.py
```diff
@@ -0,0 +1,115 @@
...
+
+        indx = get_function_index(x, y, tswitch=self.tswitch)
+
+        flst = [self.kernels[0].__call__, k_zero, k_zero, self.kernels[1].__call__]
+        K = jax.lax.switch(indx, flst, x, y)
```
daniel-dodd, line 113: Maybe on naming. I don't know what `flst` means. Maybe something like `kernel_options`. 

Also, why do we need `self.kernels[0].__call__` and `self.kernels[1].__call__`. Does this not work?
```
kernel_options = self.kernels + [k_zero].
```
Would this help with extending this kernel to more than two kernels? Or is there some issue with the computing this indexing?

gpjax/kernels/stationary/changepoint.py
```diff
@@ -0,0 +1,115 @@
...
+        for kernel in self.kernels:
+            if not isinstance(kernel, AbstractKernel):
+                raise TypeError("can only combine Kernel instances")  # pragma: no cover
+
+            if isinstance(kernel, self.__class__):
+                kernels_list.extend(kernel.kernels)
+            else:
+                kernels_list.append(kernel)
+
+        self.kernels = kernels_list
```
daniel-dodd, line 68: This is great. But we should probably through an error if we specify more than two kernels (with the current implementation)?

tests/test_kernels/test_stationary.py
```diff
@@ -241,6 +242,44 @@ class TestRationalQuadratic(BaseTestKernel):
     default_compute_engine = DenseKernelComputation()
 
 
+@pytest.mark.parametrize(
+    "kernel",
+    [RBF, RationalQuadratic, Matern12, Matern32, Matern52],
+)
+@pytest.mark.parametrize("tswitch_param", [0.1, 0.25, 0.5, 0.75, 0.9])
+def test_changepoint_kernel(kernel: AbstractKernel, tswitch_param: float) -> None:
+    n_kerns = 2
```
daniel-dodd, line 251: This looks good. The only thing is, I would like to see (a) an error is thrown if `n_kerns` is wrong and this is tested.

daniel-dodd:
Thanks @jakeyeung. Overall this looks really good. Quite excited to give this a go on some data soon. Mainly asking some clarification questions in the above review. Just the only main points is, I would suggest it is made clear that this is for 1D data. This needs to be in the docstrings and in the typing if possible. I would like to see a check for the number of kernels. If this kernel is just for two kernels (and not more), then maybe instead of `kernels` as a list, it might be more sensible to have e.g., `before_kernel`, `after_kernel`, before and after the changepoint.

gpjax/kernels/stationary/changepoint.py
jakeyeung: The idea started with implementing a basic change-point kernel similar to what GPflow has (See Change-point section in https://gpflow.github.io/GPflow/develop/notebooks/getting_started/kernels.html). 

So the location is not learned, but given as a fixed parameter. 

At the moment this piece of code only works with two kernels, so one changepoint.

I can think about converting this to allow any number of kernels n, and therefore (n-1) locations.

gpjax/kernels/stationary/changepoint.py
jakeyeung: The idea started with the GPflow documentation, but I guess if you're thinking about a paper it would be Saatci et al 2010 : (https://icml.cc/Conferences/2010/papers/170.pdf). 

I'll add this in the doc strings.

gpjax/kernels/stationary/changepoint.py
jakeyeung: They should be 1-dimensional, I've now fixed now.

gpjax/kernels/stationary/changepoint.py
jakeyeung: I've fixed this now.

gpjax/kernels/stationary/changepoint.py
```diff
@@ -0,0 +1,115 @@
...
+            """
+            # Four possible indexes: 0, 1, 2, 3
+            indx = 3  # if indx doesn't get set to 0, 1, or 2, then default 3
+            # lessthan means that at tswitch, you are already switched
+            cond1 = jnp.less(x, tswitch)
+            cond2 = jnp.less(y, tswitch)
+            indx = jnp.where(jnp.logical_and(cond1, cond2), 0, indx)
+            indx = jnp.where(jnp.logical_and(jnp.invert(cond1), cond2), 1, indx)
+            indx = jnp.where(jnp.logical_and(cond1, jnp.invert(cond2)), 2, indx)
+            return indx.squeeze().astype("uint8")
```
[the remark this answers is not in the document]
jakeyeung: The four indices can be reduced to three. I found the code easier to read to have four indices corresponding to four quadrants in a 2D matrix.

gpjax/kernels/stationary/changepoint.py
jakeyeung: I don't know the details of `jax` enough to know how the different types play with other types and conditionals. What would be a good way to test which type is best to put in here?

gpjax/kernels/stationary/changepoint.py
jakeyeung: To extend to more than two kernels, I would need to rewrite the `get_function_index()`. 

The technical part is to write a conditional operator that returns an index `indx` that can directly call `self.kernels[indx].__call__`, otherwise return `k_zero`.  I'll think about how to do this.

tests/test_kernels/test_stationary.py
jakeyeung: Sorry to ask basic question, how does write a test to make sure that a proper error is thrown?

PR #358, opened 2023-08-10
Fix bug in zero mean function and add test

Thomas-Christie:
## Type of changes

- [X] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Fixes bug whereby the constant in a zero mean function was able to change from model optimisation. 

Issue Number: **https://github.com/JaxGaussianProcesses/GPJax/issues/330**


tests/test_mean_functions.py
```diff
@@ -40,3 +51,37 @@ def test_constant(constant: Float[Array, " Q"]) -> None:
...
+    key = jr.PRNGKey(123)
+
+    x = jr.uniform(key=key, minval=0, maxval=1, shape=(20, 1))
+    y = jnp.full((20, 1), 50, dtype=jnp.float64)  # Dataset with non-zero mean
+    D = gpx.Dataset(X=x, y=y)
+
+    kernel = gpx.kernels.Constant(constant=jnp.array(0.0))
+    kernel = kernel.replace_trainable(
+        constant=False
+    )  # Prevent kernel from modelling non-zero mean
```
daniel-dodd, line 66: Is there any reason we have trainability turned off here?

tests/test_mean_functions.py
```diff
@@ -40,3 +51,37 @@ def test_constant(constant: Float[Array, " Q"]) -> None:
...
+    meanf = Zero()
+    prior = gpx.Prior(mean_function=meanf, kernel=kernel)
+    likelihood = gpx.Gaussian(num_datapoints=D.n, obs_noise=jnp.array(1e-6))
+    likelihood = likelihood.replace_trainable(obs_noise=False)
```
daniel-dodd, line 70: And trainability turned off here too?

daniel-dodd (changes requested):
Nice one @Thomas-Christie. This is looking good. 

Good idea to test training loop!

Two further tests that I'd be keen to see are (1) doing a `tree_flatten` of the `Zero` - check that that the flatten leaves are an empty list `[]`. (2) Check that you can't initialise a different value for the `constant` during class initialisation (i.e., `Zero(constant=...)` returns an Error). These would eliminate bugs, if we did any future refactoring! :)

tests/test_mean_functions.py
Thomas-Christie: I just turned off trainability in both cases such that if the `Zero` mean was still trainable, it would be the only part of the model able to be trained (and so no other aspect of the model (such as the kernel) could be used to model the non-zero mean of the data other than the `Zero` mean function).

daniel-dodd (approved):

PR #362, opened 2023-08-20
Add classes for handling posteriors during the decision making loop

Thomas-Christie:
## Type of changes

- [ ] Bug fix
- [X] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Added a `PosteriorHandler` class which can be used to obtain and update posteriors during the decision making loop as the number of datapoints in the training dataset changes (which is necessary for instantiating a likelihood object).

Also added an `AbstractPosteriorOptimizer` abstract base class for optimizing posterior hyperparameters during the decision making loop. This allows for more customisability when optimizing posteriors. Added a concrete implementation in the form of an `AdamPosteriorOptimizer` which can be used to optimize posteriors with the Adam optimizer provided by optax.

Issue Number: N/A


gpjax/decision_making/posterior_optimizer.py
```diff
@@ -0,0 +1,104 @@
...
+        Args:
+            posterior: Posterior being optimized.
+            dataset: Dataset used for optimizing posterior.
+            key: A JAX PRNG key for generating random numbers.
+        """
+        raise NotImplementedError
+
+
+@dataclass
+class AdamPosteriorOptimizer(AbstractPosteriorOptimizer):
```
thomaspinder, line 56: This feels a level of abstraction too far. What if I want to use `adamw` or `sgd`? With this abstraction I'd have subclass `AbstractPosteriorOptimizer` just to change the optimiser. Could we instead structure it as `OptaxPosteriorOptimizer` or something similar where the user supplies the relevant optimiser? It feels a bit more general.

thomaspinder: Further, what's the benefit of this object, given that it just calls `fit`?

thomaspinder (commented):
I've attempted a review. However, I'm having a pretty hard time at this point imagining how all the moving part of the decision making framework work together. A notebook or even examples in the docstrings would be really helpful. My concern is that things are becoming slightly too abstracted and taking the necessary level of control away from the user. That being said, I may be wrong, hence asking for some example(s).

gpjax/decision_making/posterior_handler.py
```diff
@@ -0,0 +1,117 @@
...
+            dataset: Dataset to get posterior for.
+            previous_posterior: Posterior being updated. This is supplied as one may
+            wish to simply increase the number of datapoints in the likelihood, without
+            optimizing the posterior hyperparameters, in which case the previous
+            posterior can be used to obtain the previously set prior hyperparameters.
+            optimize: Whether to optimize the posterior hyperparameters.
+            key: A JAX PRNG key which is used for optimizing the posterior
+            hyperparameters.
+        """
+        posterior = previous_posterior.prior * self.likelihood_builder(dataset.n)
```
henrymoss, line 109: Nice, so this steals the optimized kernel params from the last go?

gpjax/decision_making/posterior_optimizer.py
henrymoss: I agree with @thomaspinder. I think you just need to pass in optim and num_iters into the PosteriorHandler

henrymoss (approved):
LGTM. I agree that the posterior optimiser seems overkill! Impressive battery of tests though

gpjax/decision_making/posterior_optimizer.py
Thomas-Christie: Yes - I had originally kept the optimisation logic separate in anticipation of moving towards fitting models with L-BFGS further down the line. However, I think I will move this logic to the `PosteriorHandler` class, and accept that I may just have to change this class a bit once I switch to using L-BFGS.

gpjax/decision_making/posterior_handler.py
Thomas-Christie: Yes, this is included to cover for a scenario where e.g. a user wishes to run the decision making loop, but doesn't want to optimise the posterior on each iteration of the loop. With this logic they can still update the likelihood, to reflect the change in dataset size, but without changing the prior parameters.

thomaspinder (approved):

thomaspinder:
The PR looks good. However, I think #364 will affect the structure of this PR - we should be careful with this before merging into `main`.

PR #364, opened 2023-08-22
Model fitting using jaxopt solvers

frazane:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Model training using `fit` can now be called with a much wider choice of optimization algorithms provided as `jaxopt` solvers. These also include `optax` solvers (which are currently used).


gpjax/fit.py
```diff
@@ -121,48 +115,49 @@ def evaluate(self, model: LinearModel, train_data: gpx.Dataset) -> float:
...
+
+    # Initialise solver state.
+    solver.fun = _wrap_objective(solver.fun)
+    solver.__post_init__()  # needed to propagate changes to `fun` attribute
+
+    solver_state = solver.init_state(
+        model,
+        get_batch(train_data, batch_size, key) if batch_size != -1 else train_data,
+    )
+    jitted_update = jax.jit(solver.update)
```
thomaspinder, line 138: Should we have any additional unit tests to check this block is running correctly?

thomaspinder (changes requested):
Looks good @frazane! A high-level comment on the notebook updates. I see you've updated the code, but can ensure that the supporting text is aligned. In the `regression` notebook, for example, I think there should be changes

gpjax/fit.py
thomaspinder: Hmm, my concern with this approach is that it demands a lot of the user. If I've passed a bijector or a trainability status to my parameters, then I would not expect to have to explicitly apply this functionality again later on. Thoughts?

frazane: I agree with your concerns. For now I'll keep it as is and add a test to check that specific block 👍

henrymoss:
I've taken this on so we can merge in @frazane's absence. 

I've fiddled it so it all works with the new BO code and it seems to pass all the tests locally. 

I can;t work out why it wont pass on the CI though. Please can someone take a quick look at the CI logs and explain them to me @thomaspinder  or @daniel-dodd ?

daniel-dodd:
Nice work @henrymoss.

I think we need to remove the `plum` dependancy!

daniel-dodd (approved):
Have reviewed the code. Other than the aforementioned issue, this PR looks excellent to me - will give an approval ahead of fixing this, feel free to merge as and when the tests pass. Thanks, @henrymoss, @frazane. :)

henrymoss:
@daniel-dodd I actually think something is really wrong with this. 

I was updating the notebooks to use the LBFGS instead of Adam and I couldn't get good performances on the ocean or decision maker notebook. 

In the decision maker notebook, I was getting weird things like the plots of the posterior became invisible .....

henrymoss:
Also @daniel-dodd , If you use the new version of Adam (e.g. in the decision maker notebook), model fits seem much worse than with the old version of Adam

daniel-dodd:
Ah strange @henrymoss!

gpjax/fit.py
daniel-dodd, line 138: Actually would not recommend using `jit` here. The body function is traced in the `lax.scan`.

gpjax/fit.py
```diff
@@ -121,48 +115,49 @@ def evaluate(self, model: LinearModel, train_data: gpx.Dataset) -> float:
...
             batch = train_data
 
-        loss_val, loss_gradient = jax.value_and_grad(loss)(model, batch)
-        updates, opt_state = optim.update(loss_gradient, opt_state, model)
-        model = ox.apply_updates(model, updates)
-
-        carry = model, opt_state
-        return carry, loss_val
+        model, state = jitted_update(model, state, batch)
+        carry = model, state
```
daniel-dodd, line 153: `jitted_update` not necessary

docs/examples/bayesian_optimisation.py
```diff
@@ -216,10 +217,10 @@ def return_optimised_posterior(
 
     opt_posterior, history = gpx.fit(
         model=posterior,
-        objective=negative_mll,
-        train_data=data,
-        optim=ox.adam(learning_rate=0.01),
-        num_iters=1000,
+        train_data=D,
```
daniel-dodd, line 220: @henrymoss  Maybe, this changed from `train_data=data` to `train_data=D`?

henrymoss: its decision_making that is rubbish

daniel-dodd:
@henrymoss fixed it.

henrymoss:
Right then @thomaspinder  and @daniel-dodd . I have rejigged this somewhat. When I went to update some of the notebooks to use the jaxopt lbfgs, everything got worse, i.e. it was even worse than ADAM!!!

Turns out that the jaxopt lbfgs is its own implementation that lots of people say is a bit naff (similar to the torch and tensorflow versions). I have instead built a wrapper to go to Scipy (still through jaxopt), but now using Scipy's LBFGS directly. 

We now have a model fitter than works really really well and has resulted in a lot of the notebooks running a lot faster. For example, the barycentre model fits take 4 secs to get to lower loss than after 25 secs of Adam.

I had to rejig the code a little bit, because the Scipy optimisers just optimise in one go (rather than the step by step of the optax ones). This is actually great,  because it means we dont have to specify a `maxiter`, which, to me, is one of the advantages of LBFGS.

I have also strengthened the typing a bit. We now only support the optax and scipy wrappers from jaxopt. Before, the code suggested that we supported all IterativeSolvers from jaxopt, but this was nonsense. We never actually checked this and all the different solvers will likely require weird hacks to get to work.

thomaspinder:
Thanks for picking this up @henrymoss. It looks like some of the documentation is failing to build - would you mind looking into this and then, once resolved, we should be good to merge.

henrymoss:
> Thanks for picking this up @henrymoss. It looks like some of the documentation is failing to build - would you mind looking into this and then, once resolved, we should be good to merge.

I cant recreat this error on my machine :( @daniel-dodd reckons its something to do with macs....

daniel-dodd:
@henrymoss This is not what I thought earlier. I thought we were passing through the same solver to the posterior, and returning this from `fit`.

Actually what is happening is side-effect behaviour. When we call `fit` the `solver` gets modified outsider the function. This is a bug.

mblondel:
JAXopt developer here. JAXopt's LBFGS is getting better and better. We try to fix things as people report issues. The issue is that optimization with linesearch in float32 precision is very hard. Using scipy's LBFGS means that float64 precision is used, while JAX uses float32 by default. If people compare JAXopt's LBFGS and scipy's LBFGS, they should compare both with float64. Please do report issues when you encounter them.

fabianp:
Also, if you keep encountering issues in jaxopt, we would be very grateful if you can report them (ideally with a reproducible test 🙏🏼 )

adam-hartshorne:
I don't know if you have seen this,

Optimistix is a [JAX](https://github.com/google/jax) library for nonlinear solvers: root finding, minimisation, fixed points, and least squares.
https://docs.kidger.site/optimistix/

Reasons to use Optimistix:

- Optimistix is much faster to compile, and faster to run.
- Optimistix supports some solvers not found in JAXopt (e.g. [optimistix.Newton](https://docs.kidger.site/optimistix/api/root_find/#optimistix.Newton) for root-finding problems).
- Optimistix's APIs will integrate more cleanly with the scientific ecosystem being built up around [Equinox](https://github.com/patrick-kidger/equinox).
- Optimistix is much more flexible for advanced use-cases, see e.g. the way we can [mix-and-match](https://docs.kidger.site/optimistix/api/searches/introduction/) different optimisers.

https://docs.kidger.site/optimistix/faq/

PR #365, opened 2023-08-25
Add logic for running a decision making loop

Thomas-Christie:
## Type of changes

- [ ] Bug fix
- [X] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Add logic for running a decision making loop
    
Added an `AbstractDecisionMaker` abstract base class which can be
implemented in order to define a decision making loop.

A concrete implementation, in the form of a `DecisionMaker` class,
has been added. At its heart it has two core methods:
1. `ask` which is used to get a point to be queried next.
2. `tell` which is used to tell the `DecisionMaker` about new
   observations. In a typical decision making setup this will result in
   the datsets and posteriors being updated.

In addition to this, in the `DecisionMaker` a `run` method is provided,
which will automatically run the decision making loop for n steps. After
the `ask` step, the functions in the `post_ask` list will be executed,
taking as arguments the decision maker and the point chosen to be
queried next. Similarly, after the `tell` step, the functions in
the `post_tell` list are executed, taking the decision maker as the sole
argument.

Also removed the check for type-checking import warnings (TCH) from Ruff in `pyproject.toml`.

Issue Number: N/A

### DIFF
--- a/gpjax/decision_making/decision_maker.py
+++ b/gpjax/decision_making/decision_maker.py
@@ -1,0 +1,254 @@
+# Copyright 2023 The GPJax Contributors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+from abc import (
+    ABC,
+    abstractmethod,
+)
+from dataclasses import (
+    dataclass,
+    field,
+)
+
+from beartype.typing import (
+    Callable,
+    Dict,
+    List,
+    Mapping,
+)
+import jax.random as jr
+
+from gpjax.dataset import Dataset
+from gpjax.decision_making.acquisition_functions import (
+    AbstractAcquisitionFunctionBuilder,
+)
+from gpjax.decision_making.acquisition_maximizer import AbstractAcquisitionMaximizer
+from gpjax.decision_making.posterior_handler import PosteriorHandler
+from gpjax.decision_making.search_space import AbstractSearchSpace
+from gpjax.decision_making.utils import FunctionEvaluator
+from gpjax.gps import AbstractPosterior
+from gpjax.typing import (
+    Array,
+    Float,
+    KeyArray,
+)
+
+
+@dataclass
+class AbstractDecisionMaker(ABC):
+    """
+    AbstractDecisionMaker abstract base class which handles the core decision loop. The
+    decision making loop is split into two key steps, `ask` and `tell`. The `ask`
+    step is typically used to decide which point to query next. The `tell` step is
+    typically used to update models and datasets with newly queried points.
+
+    Attributes:
+        search_space (AbstractSearchSpace): Search space which is being queried
+        posterior_handlers (Dict[str, PosteriorHandler]): Dictionary of posterior
+            handlers, which are used to update posteriors throughout the decision making
+            loop. Tags are used to distinguish between posteriors. In a typical Bayesian
+            optimisation setup one of the tags will be `OBJECTIVE`, defined in
+            decision_making.utils.
+        datasets (Dict[str, Dataset]): Dictionary of datasets, which are augmented with
+            observations throughout the decision making loop. In a typical setup they are
+            also used to fit the posteriors, using the `posterior_handlers`. Tags are used
+            to distinguish datasets, and correspond to tags in `posterior_handlers`.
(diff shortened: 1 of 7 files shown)


thomaspinder (approved):

gpjax/decision_making/decision_maker.py
```diff
@@ -0,0 +1,254 @@
...
+    Array,
+    Float,
+    KeyArray,
+)
+
+
+@dataclass
+class AbstractDecisionMaker(ABC):
+    """
+    AbstractDecisionMaker abstract base class which handles the core decision loop. The
```
henrymoss, line 51: maybe say what "core decision loop" means. I think its important to stress the sequential aspect

gpjax/decision_making/decision_maker.py
```diff
@@ -0,0 +1,254 @@
...
+    decision making loop is split into two key steps, `ask` and `tell`. The `ask`
+    step is typically used to decide which point to query next. The `tell` step is
+    typically used to update models and datasets with newly queried points.
+
+    Attributes:
+        search_space (AbstractSearchSpace): Search space which is being queried
```
henrymoss, line 57: search space over which we can evaluate

gpjax/decision_making/decision_maker.py
```diff
@@ -0,0 +1,254 @@
...
+        posterior_handlers (Dict[str, PosteriorHandler]): Dictionary of posterior
+            handlers, which are used to update posteriors throughout the decision making
```
henrymoss, line 59: maybe say that what we are calling posteriors would often usually be referred to as models in the usual model-based decision making literature

gpjax/decision_making/decision_maker.py
```diff
@@ -0,0 +1,254 @@
...
+            loop. Tags are used to distinguish between posteriors. In a typical Bayesian
+            optimisation setup one of the tags will be `OBJECTIVE`, defined in
+            decision_making.utils.
+        datasets (Dict[str, Dataset]): Dictionary of datasets, which are augmented with
+            observations throughout the decision making loop. In a typical setup they are
+            also used to fit the posteriors, using the `posterior_handlers`. Tags are used
```
henrymoss, line 65: *update* the posterior. Dont let CH see you saying "fit" posterior. No NNs here!

gpjax/decision_making/decision_maker.py
```diff
@@ -0,0 +1,254 @@
...
+            observation_datasets (Mapping[str, Dataset]): Dictionary of datasets
+                containing new observations. Tags are used to distinguish datasets, and
+                correspond to tags in `posterior_handlers` in a typical setup.
+            key (KeyArray): JAX PRNG key for controlling random state.
+        """
+        raise NotImplementedError
+
+
+@dataclass
+class DecisionMaker(AbstractDecisionMaker):
```
henrymoss, line 124: I wonder if the more abstract one should not have acquisition functions or optimizers in?

henrymoss: perhaps then this class could be utility-based decision maker or something?

henrymoss: as the difference between the two does seem to be mainly about having an acq to play with (and plot)

henrymoss: We could then have a RS one subclassed from abstract without a need for defining an acq

gpjax/decision_making/decision_maker.py
```diff
@@ -0,0 +1,254 @@
...
+        for tag, observation_dataset in observation_datasets.items():
+            self.datasets[tag] += observation_dataset
+
+        for tag, posterior_handler in self.posterior_handlers.items():
+            key, _ = jr.split(key)
+            self.posteriors[tag] = posterior_handler.update_posterior(
+                self.datasets[tag], self.posteriors[tag], optimize=True, key=key
+            )
+
+    def run(
```
henrymoss, line 214: could this be a method of the more abstract class as  well? It just requires ask and tell right?

henrymoss (approved):
Looks great. Just a couple Qs about organisation here

gpjax/decision_making/decision_maker.py
```diff
@@ -0,0 +1,254 @@
...
+    adds a newly queried point to the `datasets` and updates the `posteriors`.
+
+    This can be run as a typical ask-tell loop, or the `run` method can be used to run
+    the decision making loop for a fixed number of steps. Moreover, the `run` method executes
+    the functions in `post_ask` and `post_tell` after each ask and tell step
+    respectively. This enables the user to add custom functionality, such as the ability
+    to plot values of interest during the optimization process.
+    """
+
+    def __post_init__(self):
```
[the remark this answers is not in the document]
Thomas-Christie: Hadn't included originally in case people wanted a decision maker which didn’t optimise all the models/there was some case where people didn’t need to have exactly the same datasets as models. But I suppose in this case they can just override `__post_init__`, so will move to the abstract class.

gpjax/decision_making/decision_maker.py
Thomas-Christie: Yeah, will just add to abstract class. I'll also add a concrete implementation of `tell` to the abstract class as this should be pretty standard across different decision-making methods.

PR #385, opened 2023-09-06
Prepare for future batching

Thomas-Christie:
## Type of changes

- [ ] Bug fix
- [X] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

In the future certain utility functions will be "batched" utility functions which quantify the utility of querying a *batch* of points rather than a single point. These utility functions may in turn be optimised differently.

Therefore, I propose we have a `SinglePointUtilityFunction` type for non-batched utility functions, which has type annotation `[N  D] -> [N 1]`, and in the future we may add `BatchUtilityFunction` which has type  annotation `[N B D] -> [N 1]`. Distinguishing between these makes it easy for users to see which utility functions support batching.

Also refactored UtilityDrivenDecisionMaker to enable for Thompson sampling to be used in a batched context, and updated unit tests to reflect this. Finally, added a new notebook to serve as an introduction to the `decision_making` module.

Issue Number: N/A

### DIFF
--- a/gpjax/decision_making/decision_maker.py
+++ b/gpjax/decision_making/decision_maker.py
@@ -16,6 +16,7 @@ from abc import (
     ABC,
     abstractmethod,
 )
+import copy
 from dataclasses import dataclass
 
 from beartype.typing import (
@@ -24,12 +25,16 @@ from beartype.typing import (
     List,
     Mapping,
 )
+import jax.numpy as jnp
 import jax.random as jr
 
 from gpjax.dataset import Dataset
 from gpjax.decision_making.posterior_handler import PosteriorHandler
 from gpjax.decision_making.search_space import AbstractSearchSpace
-from gpjax.decision_making.utility_functions import AbstractUtilityFunctionBuilder
+from gpjax.decision_making.utility_functions import (
+    AbstractUtilityFunctionBuilder,
+    ThompsonSampling,
+)
 from gpjax.decision_making.utility_maximizer import AbstractUtilityMaximizer
 from gpjax.decision_making.utils import FunctionEvaluator
 from gpjax.gps import AbstractPosterior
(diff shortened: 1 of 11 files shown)


thomaspinder (approved):

gpjax/decision_making/decision_maker.py
```diff
@@ -222,28 +240,64 @@ class UtilityDrivenDecisionMaker(AbstractDecisionMaker):
...
+                maximizer = self.utility_maximizer.maximize(
+                    decision_function, self.search_space, key
+                )
+                maximizers.append(maximizer)
+                _, key = jr.split(key)
+
+            maximizers = jnp.concatenate(maximizers)
+            return maximizers
+        else:
+            raise NotImplementedError(
```
henrymoss, line 301: this keeps it tidy for now

gpjax/decision_making/decision_maker.py
```diff
@@ -222,28 +240,64 @@ class UtilityDrivenDecisionMaker(AbstractDecisionMaker):
...
+            Float[Array, "B D"]: Point(s) to be queried next.
         """
-        self.current_utility_function = (
-            self.utility_function_builder.build_utility_function(
-                self.posteriors, self.datasets, key
+        self.current_utility_functions = []
+        maximizers = []
+        if isinstance(self.utility_function_builder, ThompsonSampling) or (
+            (not isinstance(self.utility_function_builder, ThompsonSampling))
+            and (self.batch_size == 1)
```
henrymoss, line 279: does this still not work with batch_size=1? This logic is a little tricky to understand

henrymoss (approved):
LGTM

gpjax/decision_making/decision_maker.py
Thomas-Christie: It's restricting batch size to be 1 for non-TS acquisition functions if someone happens to implement one for now.

henrymoss: got it. Maybe you can make the logic clearer so people like me dont get confused?

PR #400, opened 2023-10-06
Fix obs_noise confusion

henrymoss:
I have changed all uses of obs_noise (the noise level of our Gaussian likelihood) to obs_stddev which is much clearer.

We were previously inconsistent with this (sometimes it was var and sometimes it was sd) which actually lead to some bugs which have been fixed in this PR.

I have changed a couple of obs_noise=1e-6  to be obs_stddev=1e-3 to keep the same behaviour in our notebooks.


gpjax/variational_families.py
```diff
@@ -683,7 +683,7 @@ def predict(
         Lz_inv_Kzx = cola.solve(Lz, Kzx)
 
         # A = Lz⁻¹ Kzt / o
-        A = Lz_inv_Kzx / jnp.sqrt(noise)
+        A = Lz_inv_Kzx / jnp.sqrt(noise_var)
```
daniel-dodd, line 686: Here do we need the `jnp.sqrt`? Could we just use `obs_stddev`?

gpjax/variational_families.py
```diff
@@ -683,7 +683,7 @@ def predict(
         Lz_inv_Kzx = cola.solve(Lz, Kzx)
 
         # A = Lz⁻¹ Kzt / o
```
daniel-dodd, line 685: Think pre-commit stuff has gone rogue 😱. The comment should has a `σ` not an `o`.

daniel-dodd (approved):
Thanks @henrymoss. This looks really good to me. `likelihood.obs_stddev` is fine, but maybe  something like e.g., `likelihood.root_variance` might read more cleanly. But I leave this up to you. :)

thomaspinder (approved):
Looks good. Thanks @henrymoss. Just need to fix Dan's comment re. the diagonal noise component and it can be merged.

PR #408, opened 2023-11-06
Namespace cleanup

thomaspinder:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Tests
- [X] Other

## Checklist

## Description

This PR cleans up the package's namespace, as per Slack discussions. 

Issue Number: N/A

### DIFF
--- a/gpjax/__init__.py
+++ b/gpjax/__init__.py
@@ -13,120 +13,40 @@
 # limitations under the License.
 # ==============================================================================
 from gpjax import (
+    base,
     decision_making,
+    gps,
     integrators,
+    kernels,
+    likelihoods,
+    mean_functions,
+    objectives,
+    sklearn,
+    variational_families,
 )
-from gpjax.base import (
-    Module,
-    param_field,
-)
-from gpjax.citation import cite
 from gpjax.dataset import Dataset
+from gpjax.citation import cite
 from gpjax.fit import fit
-from gpjax.gps import (
-    Prior,
-    construct_posterior,
-)
-from gpjax.kernels import (
-    RBF,
-    RFF,
-    AbstractKernel,
-    BasisFunctionComputation,
-    CatKernel,
-    ConstantDiagonalKernelComputation,
-    DenseKernelComputation,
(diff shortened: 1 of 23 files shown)


gpjax/__init__.py
```diff
@@ -13,120 +13,37 @@
...
-    VariationalGaussian,
-    WhitenedVariationalGaussian,
-)
 
 __license__ = "MIT"
 __description__ = "Didactic Gaussian processes in JAX"
 __url__ = "https://github.com/JaxGaussianProcesses/GPJax"
 __contributors__ = "https://github.com/JaxGaussianProcesses/GPJax/graphs/contributors"
-__version__ = "0.7.1"
+__version__ = "0.7.2"
```
st--, line 34: Technically this is a breaking change (`gpx.Matern52` will no longer work as before), so should be
```suggestion
__version__ = "0.8.0"
```
if you want to follow semantic versioning (which I would be strongly in favour of).

gpjax/__init__.py
```diff
@@ -13,120 +13,37 @@
 # limitations under the License.
 # ==============================================================================
 from gpjax import (
+    base,
     decision_making,
+    gps,
     integrators,
-)
-from gpjax.base import (
-    Module,
```
st--, line 20: It's probably better to be consistent but `gpjax.Module` (and perhaps the `param_field`) is the one thing I would've just kept in the "top-level" namespace.

st-- (approved):
Looks good to me, though I would
- suggest (but not bother arguing about) keeping `Module` and `param_field` in the top-level name-space
- strongly suggest (and be willing to argue) that this is breaking and bump the version to 0.8.0 to follow semantic versioning
- there's pending merge conflicts that I've not reviewed

gpjax/__init__.py
thomaspinder: Resolved in forthcoming commit

gpjax/__init__.py
thomaspinder: Agreed

PR #410, opened 2023-11-08
fix config import

henrymoss:
Updated how we import config from jax. Our current method has been deprecated on the most recent version of jax. The new method works for the older versions and the new versions.


daniel-dodd (approved):
Sweet.

PR #417, opened 2023-11-26
Fix typo in CollapsedELBO

meta-inf:
## Type of changes

- [x] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

I want to fix a typo in `CollapsedELBO` where noise standard deviation was incorrectly used in place of variance; compare the occurrence of `noise` and the comments around the change.  You can also verify only the new version is correct using the following code:

```python
import gpjax as gpx, jax, jax.numpy as jnp
jax.config.update('jax_enable_x64', True)

key1, key2 = jax.random.split(jax.random.PRNGKey(23))
X = jax.random.normal(key1, shape=(10, 1))
X1 = jax.random.normal(key2, shape=(5, 1))
prior = gpx.Prior(
    mean_function=gpx.mean_functions.Zero(),
    kernel=gpx.kernels.Matern32(lengthscale=0.2, variance=1.))
Y = jax.vmap(prior.sample_approx(1, key2))(X)[:, 0]
likelihood = gpx.Gaussian(num_datapoints=X.shape[0], obs_stddev=3)
posterior = prior * likelihood

dat = gpx.Dataset(X, Y, None)
print(gpx.ConjugateMLL().step(posterior, dat))

# use all training data as inducing point should lead to ELBO == log marginal likelihood
full_variational = gpx.CollapsedVariationalGaussian(posterior, X)
print(gpx.CollapsedELBO().step(full_variational, dat))
```

### DIFF
--- a/tests/test_objectives.py
+++ b/tests/test_objectives.py
@@ -175,7 +175,12 @@ def test_collapsed_elbo(
     assert isinstance(evaluation, jax.Array)
     assert evaluation.shape == ()
 
-    # with pytest.raises(TypeError):
+    # Data on the full dataset should be the same as the marginal likelihood
+    q = gpx.CollapsedVariationalGaussian(posterior=p * likelihood, inducing_inputs=D.X)
+    mll = ConjugateMLL(negative=negative)
+    expected_value = mll(p * likelihood, D)
+    actual_value = negative_elbo(q, D)
+    assert jnp.abs(actual_value - expected_value) / expected_value < 1e-6
 
 
 @pytest.mark.parametrize("num_datapoints", [1, 2, 10])


daniel-dodd (approved):
Thanks @meta-inf. Prior to merging, I think it would be good to use your example as a test to prevent this from reoccurring. If you don't get to it, that's fine, I can add this later in the week.

meta-inf:
That's great, thanks for getting back.  I'm not sure how you want to structure the test, so perhaps you could look into this when you have time.

daniel-dodd:
No worries @meta-inf. Have just added a simple relative error test for the equivalence of the mll and collapsed elbo on full dataset. Thanks again for spotting this, will merge, soon as tests pass! 🤞

PR #419, opened 2023-11-28
Cleanup src

thomaspinder:
## Type of changes

- [X] Bug fix
- [ ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Tighten up the underlying codebase by removing aspects of the code that are either incongruent with the entire API, or introduce too much boilerplate code for their returned utility. The code being removed is deeply woven into the source code, so adding deprecations is not possible.

Notes:
- Once approved, I'll open an issue to refactor the pathwise sampling notebook with masking removed.
- The `strict=True` argument has been added to all `zip` calls to make Ruff compliant with Python 3.10 standards. See [B905](https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/)

Issue Number: N/A


docs/examples/graph_kernels.py
```diff
@@ -158,6 +174,7 @@
     model=posterior,
     objective=jit(gpx.ConjugateMLL(negative=True)),
     train_data=D,
+    max_iters=1000
```
henrymoss, line 177: did this fix it? Seems like a huge number for LBFGS

gpjax/gps.py
```diff
@@ -495,66 +478,41 @@ def predict(
...
+        covariance = Ktt - jnp.matmul(Kxt.T, Sigma_inv_Kxt)
         covariance += cola.ops.I_like(covariance) * self.prior.jitter
         covariance = cola.PSD(covariance)
-        rval = GaussianDistribution(jnp.atleast_1d(mean.squeeze()), covariance)
-        if m == 1:
-            return rval
-        else:
-            return ReshapedDistribution(rval, (n_test, m))
+
+        return GaussianDistribution(jnp.atleast_1d(mean.squeeze()), covariance)
```
henrymoss, line 515: Readable!

henrymoss (approved):
Nice bit of surgery Dr. Pinder

docs/examples/graph_kernels.py
thomaspinder: No - the error persists. Works locally

PR #425, opened 2023-11-30
Add generic typing information for gps objects

frazane:
## Type of changes

- [ ] Bug fix
- [x] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

This PR improves the static typing information using generic types. For now I only focused on `Prior` and `Posterior` objects since they are the most important. I also tried to work out how to do it for kernels (with `compute_engine` as a generic) but the fact that we use a default with dataclasses is making it very hard (see https://peps.python.org/pep-0696/).


thomaspinder:
What's happened to codecov?!

thomaspinder (approved):

PR #426, opened 2023-12-02
Perhaps finally a decent LBFGS?

henrymoss:
I stole @daniel-dodd "back-of-the-envolope" implementation of a direct scipy minimize interface and fiddled it a bit.

It seems to work so much better (i.e. can optimize @thomaspinder's graph kernel).

Also means we can remove the jaxopt dependency

### DIFF
--- a/docs/examples/graph_kernels.py
+++ b/docs/examples/graph_kernels.py
@@ -154,14 +154,11 @@ print(gpx.cite(kernel))
 # With a posterior defined, we can now optimise the model's hyperparameters.
 
 # %%
-opt_posterior, training_history = gpx.fit(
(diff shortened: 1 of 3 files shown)


henrymoss:
Something a bit weird happened when I updated my lock though!

thomaspinder:
Looks like something has broken with the decoupled sampler?

henrymoss:
> Looks like something has broken with the decoupled sampler?

I managed to change the beartype version I think (when I deleted jaxopt). Fiddled the poetry files back to normal now

thomaspinder:
I'm not 100% sure what's going on here. Why are you running poetry update though?

thomaspinder (approved):

PR #452, opened 2024-06-05
Adding tagged parameters and updated notebooks

thomaspinder:
## Type of changes

- [ ] Bug fix
- [X ] New feature
- [ ] Tests
- [ ] Other

## Checklist

## Description

Please describe your changes here. If this fixes a bug, please link to the issue, if possible.

Issue Number: N/A

### DIFF
--- a/gpjax/kernels/non_euclidean/graph.py
+++ b/gpjax/kernels/non_euclidean/graph.py
@@ -106,7 +106,7 @@ class GraphKernel(StationaryKernel):
         *,
         S,
         **kwargs,
-    ) -> Float[Array, ""]:
+    ):
         Kxx = (jax_gather_nd(self.eigenvectors.value, x) * S.squeeze()) @ jnp.transpose(
             jax_gather_nd(self.eigenvectors.value, y)
         )  # shape (n,n)
--- a/gpjax/parameters.py
+++ b/gpjax/parameters.py
@@ -7,11 +7,14 @@ from jax.typing import ArrayLike
 import tensorflow_probability.substrates.jax.bijectors as tfb
 
 T = tp.TypeVar("T", bound=tp.Union[ArrayLike, list[float]])
+ParameterTag = tp.Union[
+    str, tp.Literal["positive", "real", "sigmoid", "lower_triangular", "static"]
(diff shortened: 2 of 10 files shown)


gpjax/parameters.py
frazane, line 1: I thought we were going to get rid of classes since having both is kind of redundant. But then I see it can be useful to have pre-defined classes for cases with multiple tags! 👍 Except...can we have multiple tags? 

The only thing I don't like about handling transformations this way is that we have to define multiple tags for the same bijectors if the parameterization of the bijector changes, for example:
```python
"positive": tfb.Softplus()
"greater_than_one": tfb.Softplus(low=1.0)
"greater_than_two": tfb.Softplus(low=2.0)
```
but this is possibly an edge case, I am not sure.

pyproject.toml
```diff
@@ -19,7 +19,7 @@ packages = [{ include = "gpjax" }]
 python = ">=3.10,<3.12"
 jax = ">=0.4.10"
 jaxlib = ">=0.4.10"
-flax = {git = "https://github.com/google/flax.git"}
+flax = { git = "https://github.com/google/flax.git" }
```
frazane, line 22: We could use the pipy distribution at this point

gpjax/parameters.py
```diff
@@ -7,11 +7,14 @@
 import tensorflow_probability.substrates.jax.bijectors as tfb
 
 T = tp.TypeVar("T", bound=tp.Union[ArrayLike, list[float]])
+ParameterTag = tp.Union[
+    str, tp.Literal["positive", "real", "sigmoid", "lower_triangular", "static"]
+]
 
 
 def transform(
```
frazane, line 15: Will this work for multiple tags?

gpjax/kernels/non_euclidean/graph.py
```diff
@@ -106,7 +106,7 @@ def __call__(  # TODO not consistent with general kernel interface
         *,
         S,
         **kwargs,
-    ) -> Float[Array, ""]:
```
frazane, line 109: I think this should be "N N"

PR #455, opened 2024-06-25
Added two new tutorial files (without interactive matplotlib, so it can pass tests)

neel-maniar:
## Type of changes

- [ ] Bug fix
- [ ] New feature
- [x] Documentation / docstrings
- [ ] Tests
- [ ] Other

## Checklist

## Description

These two notebooks follow the work of the paper on Linearly Constrained Gaussian Processes (see citation in notebooks).

They demonstrate 
- a way to encode two types of linear constraints into the kernel of a GP
- mathematical explanations which follow the notation of the paper
- another example of implementing Multi-Output GPs in GPJax.

Issue Number: N/A

### DIFF
--- a/docs/examples/2D_div_free.py
+++ b/docs/examples/2D_div_free.py
@@ -1,0 +1,417 @@
+# ---
+# jupyter:
+#   jupytext:
+#     text_representation:
+#       extension: .py
+#       format_name: percent
+#       format_version: '1.3'
+#       jupytext_version: 1.16.1
+#   kernelspec:
+#     display_name: project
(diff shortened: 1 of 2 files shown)


thomaspinder:
Hi @neel-maniar - thanks for this PR, the notebooks look super interesting. We've just released a large improvement to VTLab in v0.9. Most of the breaking changes are hidden away behind the scenes; however, I think they may affects aspects of your notebooks. I'll checkout the PR locally in the coming days and provide pointer on which lines may need updating and how this can be done.

PR #494, opened 2025-03-03
Fix tests with jax 0.5.1

stephen-huan:
## Checklist

## Description

Fix some failing tests with jax 0.5.1. It seems from 0.5.0 -> 0.5.1 the numerics got slightly worse (especially for inverse).

For the `get_batch` test, as I understand since it's sampling with replacement it's possible to get unlucky and have two rows coincide. This is most likely when the batch size is large (the tests that fail are when the batch size is the same as the dataset size). We'd expect `n_dim * batch_size / n_data` elements to collide (each row has a `1 / n_data` chance of colliding, each collision makes `n_dim` elements the same, and there are `batch_size` rows).

Related to https://github.com/JaxGaussianProcesses/GPJax/issues/490.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -29,10 +29,10 @@ jobs:
       - name: Install Hatch
         uses: pypa/hatch@install
 
-      # # Install the dependencies
-      # - name: Check docstrings
-      #   run: |
-      #     hatch run dev:docstrings
+      # Install the dependencies
+      - name: Check docstrings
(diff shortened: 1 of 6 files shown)


stephen-huan:
CI failures seem unrelated.

thomaspinder:
Yes, unclear what is causing the CI failures right now. I am unable to reproduce locally.

thomaspinder (approved):

Issue #490, opened 2025-01-20
dev: Support for JAX >= 0.4.28


theo-brown:
Hi all,
Are there plans to support newer versions of JAX? 
A few of the other projects I work on have moved to 0.4.35, which has some breaking changes.

theo-brown:
Sure, I'll have a look at it tomorrow.

theo-brown:
Unpinning `jax` and `jaxlib` also requires unpinning `cola-ml`, as `jax==0.4.30` is only compatible with `cola-ml>=0.0.6` (see [issue #93](https://github.com/wilson-labs/cola/pull/93) and [release notes v0.0.6](https://github.com/wilson-labs/cola/releases/tag/v0.0.6) on the cola repo.)

There's an open PR ([#113](https://github.com/wilson-labs/cola/pull/113) on the cola repo) that cola needs to merge to support `jax>=0.4.35` (based on the removal of `jax.experimental.host_callback`) and `jax>=0.5.0` (based on the changes to the PRNG). I've pinged the developers to try and move this along - it's been waiting for ~5 weeks.

Once that's done, we will need to regenerate some of the test data for GPJax, as the generated numbers will be different.

stephen-huan:
I've been [packaging gpjax](https://github.com/stephen-huan/maipkgs/blob/23c47507fe46e0b86eef709c4bf07fd24cc187bb/pkgs/gpjax/default.nix) with nix and running the tests. I don't think I patched anything other than https://github.com/wilson-labs/cola/pull/113. Also https://github.com/google/jaxopt/pull/591 needs to be fixed; this raises a `SyntaxError` when running the tests and not a `SyntaxWarning`.

```
ImportError while loading conftest '.../GPJax/tests/conftest.py'.
tests/conftest.py:8: in <module>
    import gpjax  # noqa: F401
.venv/lib64/python3.12/site-packages/jaxtyping/_import_hook.py:230: in exec_module
    return super().exec_module(module)
gpjax/__init__.py:21: in <module>
    from gpjax import (
.venv/lib64/python3.12/site-packages/jaxtyping/_import_hook.py:230: in exec_module
    return super().exec_module(module)
gpjax/decision_making/__init__.py:15: in <module>
    from gpjax.decision_making.decision_maker import (
.venv/lib64/python3.12/site-packages/jaxtyping/_import_hook.py:230: in exec_module
    return super().exec_module(module)
[...]

PR #497, opened 2025-03-24
Unpin JAX version; add patch to specific cola version

theo-brown:
## Checklist

*Changes only to pyproject.toml, so these do not apply*

## Description

- Added a pin to a specific patch of cola-ml, pending it being made part of a release.
- Unpinned JAX, JAXlib, jaxopt, flax

Issue Number: #490


thomaspinder:
Can you try rebasing from #498 and see if this fixes the issues please?

theo-brown:
Oops, my unpin of JAX might be an overreach - should I limit this PR to just unpinning cola?

thomaspinder:
Yes, let's restrict this just to cola right now. Once we've removed jaxopt as a dependency, then I think we're good to remove the JAX pin and release a new version :)

theo-brown:
Rebased and reduced to only pinning cola.

7 tests fail with `AttributeError: 'jaxlib.xla_extension.DeviceList' object has no attribute 'device'. Did you mean: 'devices'?`, which I presume is a JAX version compatibility problem. I'm not super worried about chasing this down exactly, because it feels like everything's a bit interdependent.

Combining the cola pin with unpinning jax, jaxlib, and flax (required to match unpinned jax version) gives 100% test success.

Thomas-Christie:
JFYI I noticed there was a [new release](https://github.com/wilson-labs/cola/releases/tag/v0.0.7) of Cola yesterday, so we can pin to that now instead of the specific commit.

thomaspinder:
Thanks for the effort here, but I'm closing as the issue is resolve in the latest release.

PR #512, opened 2025-05-03
Introduce NonNegativeReal parameters

jamesbriant:
## Checklist

- [ ] ~~I've added docstrings for the new code.~~ N/A

## Description

Introduced a new NonNegativeReal parameter type. This has two primary applications.

### Kernel variance parameter

This has been implemented into kernel variance parameters. This allows kernels to be 'turned off' as required by setting `variance=gpx.parameter.Static(0.0)`. This is potentially useful for kernel searching algorithms. 

I have been using this feature as an optional add-on kernel which defaults to 'off'.

### Gaussian likelihood `obs_stddev`

By allowing `gpx.likelihoods.Gaussian()` to accept `obs_stddev=0.0` we can 'turn off' the likelihood. This is important for GP emulators of computer models. The easiest way to see this working well is by removing the observation noise from the regression example on the documentation.

There is perhaps a cleaner solution by defining a new likelihood class, `emulator` say, with no `obs_stddev` argument which simply returns the required number of zeros when called.

#### Why is this needed at all?

One could argue that this is not necessary as using `gps.Prior()` should be sufficient for building emulators. From a development point of view, it is easier to build tools that rely purely on the `AbstractPosterior` class with a likelihood of fixed structure baked in rather than also having to account for the `Prior` class thus rendering a likelihood 'optional'. Hence a quick 'hack' such as my implementation, or potentially an arguably cleaner `emulator` likelihood is easier to account for.

Issue Number: N/A

### DIFF
--- a/examples/constructing_new_kernels.py
+++ b/examples/constructing_new_kernels.py
@@ -33,7 +33,6 @@ from jaxtyping import (
     install_import_hook,
 )
 import matplotlib.pyplot as plt
-import numpyro.distributions as npd
 from numpyro.distributions import constraints
 import numpyro.distributions.transforms as npt
 
@@ -51,8 +50,6 @@ config.update("jax_enable_x64", True)
 with install_import_hook("gpjax", "beartype.beartype"):
     import gpjax as gpx
 
-
-tfb = tfp.bijectors
 
 # set the default style for plotting
 use_mpl_style()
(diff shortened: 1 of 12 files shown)


thomaspinder:
Thanks @jamesbriant - let me begin taking a look

PR #514, opened 2025-05-19
Add `fit_lbfgs` using Optax

theo-brown:
## Checklist

## Description

L-BFGS tends to be the most commonly used optimizer for fitting GPs (used, for example, in GPytorch/Botorch).

`gpx.fit_scipy` under the hood used `scipy.optimize.minimize`, which defaults to running L-BFGS (note, not `jax.scipy` but the original scipy, as `jax.scipy` only has BFGS rather than the reduced memory L-BFGS).
Optax has an [LBFGS implementation](https://optax.readthedocs.io/en/stable/_collections/examples/lbfgs.html#l-bfgs-as-a-solver), but **the API requires some additional arguments, namely `value`, `grad`, and `value_fn` in the `optim.update` step, that means that it can't be directly with `gpx.fit`**. Hence, in this PR I've added a new function, `fit_lbfgs`, that's a hybrid between the `fit_scipy` and `fit` functions, using Optax's LBFGS solver in a lax.while loop.

## Notes

- There are still potential use cases where `fit_scipy` might be preferable. For example, on a CPU-only system with a small number of datapoints, `fit_scipy` is in fact faster than `fit_lbfgs`, as the compilation time of the lax loop is nontrivial. However, **for larger numbers of datapoints or iterations, `fit_lbfgs` is faster, even on CPU.** 
- I chose to use a `lax.while` rather than a `lax.scan` because it allows early termination. I deemed this to be more important than logging the full history of the objective values. It also more closely reflects the `fit_scipy` interface.

Happy to be corrected on any of the above! Let me know what you think.

### DIFF
--- a/gpjax/__init__.py
+++ b/gpjax/__init__.py
@@ -32,6 +32,7 @@ from gpjax.citation import cite
 from gpjax.dataset import Dataset
 from gpjax.fit import (
     fit,
+    fit_lbfgs,
     fit_scipy,
 )
 
@@ -56,5 +57,6 @@ __all__ = [
     "fit",
     "Module",
     "param_field",
+    "fit_lbfgs",
     "fit_scipy",
 ]
--- a/gpjax/fit.py
+++ b/gpjax/fit.py
@@ -15,13 +15,13 @@
 
 import typing as tp
 
-from flax import nnx
 import jax
-from jax.flatten_util import ravel_pytree
 import jax.numpy as jnp
 import jax.random as jr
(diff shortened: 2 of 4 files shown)


theo-brown:
If it would be helpful/interesting, I can provide a speed comparison between `fit_scipy` and `fit_lbfgs`.

thomaspinder:
Thanks for this @theo-brown - this is something I’d love to get merged. Can you take a look at the failing test please?

theo-brown:
Oops, good catch! Fixed :)

Issue #515, opened 2025-05-19
feat: Adding prior regularisation to marginal likelihood fit


theo-brown:
# Feature Request

It is often useful to regularise the marginal likelihood fit of a GP with additional information known a priori. For example, if I know my data is very smooth but very noisy, I might want to force my fit routine to prefer solutions with large σ and large ℓ, rather than going for very small ℓ and σ (overfitting to the noise). This is particularly important in high dimensions.

The 'exact' way of doing this is fully Bayesian inference (e.g. with NUTS), which is often untenable. Instead, other GP libraries (e.g. [GPytorch](https://github.com/cornellius-gp/gpytorch/blob/b017b9c3fe4de526f7a2243ce12ce2305862c90b/gpytorch/mlls/exact_marginal_log_likelihood.py#L83)) **add the log-probability of the hyperparameters under the prior to the marginal likelihood.**

In GPJax, this could look something like:

1. Adding an optional `prior` argument to `gpx.parameters.Parameter`
2. Adding a tree map call in `gpx.objectives.conjugate_mll` that computes the log prob of all the parameters under their respective priors, if defined
3. Adds the log probs to the mll

**If the feature request is approved, would you be willing to submit a PR?**
Yes!

thomaspinder:
Yes! I would absolutely love to see GPJax support prior specification on hyperparameters. I think your high-level approach is correct, and I think integration with numpyro is possible. For sure, if we could integrate, then it unlocks a lot of utility in GPJax by enabling composition with complete Bayesian modelling workflows. Perhaps you'd like to look at the open PR #510 that @juanitorduz and I have been exploring.

TobyBoyne:
Hi @theo-brown and @thomaspinder,
Thanks both for working on this! Has there been any further work into this? From what I can tell, the PR was written before a switch to Equinox, so I think the PR would need to be rewritten with this in mind?

If neither of you are able to work on this, and it's still something that you want in GPJax, I'm very happy to submit a PR of my own. It doesn't seem to me that you can attach priors to `paramax` parameters, but I'm sure there are other possible solutions (eg. maintaining a separate PyTree of priors whose structure matches the GP model).

Thanks! Looking forward to hearing your thoughts.

thomaspinder:
[...]

PR #516, opened 2025-05-19
Add prior regularisation to `conjugate_mll`

theo-brown:
## Checklist

## Description

Added prior regularisation to `conjugate_mll`.

Issue Number: #515


theo-brown:
Hmm, it looks like 
```
jax.tree.map(
    lambda x: x._prior,
    posterior,
    is_leaf=lambda x: isinstance(x, gpx.parameters.Parameter)
    and not isinstance(x, gpx.parameters.Static),
)
```
is unable to descend into `ConjugatePosterior`, perhaps because it's not a Pytree? See discussion #513.

theo-brown:
@thomaspinder let me know how this sounds to you. Happy to add some extra tests, such as checking that it works alright with the fit routine, if that would be helpful. Wanted to make sure I wasn't contributing to test bloat!

thomaspinder:
I don't think there's such a thing as test bloat here - I would like to see some more tests and documentation. For example, are we assuming that the user is supplying parameters on the constrained parameter value? We optimise with respect to the unconstrained parameter though, so I think we might be missing a log-determinant of the Jacobian term when evaluating the log-prior density?

thomaspinder:
Thanks for seeking clarification upstream in Flax also. Perhaps related to #513 - it’s very possible that we’re doing something fundamentally wrong in our handling of state. I’ll be interested to hear from Flax if there’s a more efficient way for us to work.

thomaspinder:
I see the upstream issue in Flax is resolved. Do you need any further support here, or do you think it'd be best to first address #513 ?

theo-brown:
Time is the limiting factor! I think because this will be easier to resolve I'll work on this before #513.

Issue #441, opened 2024-03-08
Conflict between plum-dispatch and cola-plum-dispatch


vabor112:
# Bug Report

**GPJax version: 0.8.0**

**Current behavior:**

I am trying to update the [existing integration](https://github.com/GPflow/GeometricKernels/blob/9de3a71fda929b741dfacf5925472599d6f02b81/notebooks/frontends/GPJax.ipynb) of [GeometricKernels](https://github.com/GPflow/GeometricKernels/pull/110) with GPJax so that it works with newer versions of GPJax. It works okay for GPJax 0.6.9. However, for the current GPJax 0.8.0, I hit two problems.

The first one is exactly #397, which, although quite annoying, can be fixed by downgrading `tensorflow` to version 2.13.

The second one is illustrated in the **Related code** section below. I believe it is concenred with plum-dispatch, which we use extensively in GeometricKernels to support multiple backends. GPJax uses [cola](https://github.com/wilson-labs/cola) which in its turn relies on a fork of `cola`, [cola-plum-dispatch](https://github.com/mfinzi/plum). This unmaintained fork uses the same namespace `plum` (which seems like a terrible sin) and gets overriden by the actual `plum` that GeometricKernels uses, causing the error below. I believe this is similar to this [issue](https://github.com/JiaYaobo/fenbux/issues/1).

**Expected behavior:**

I am not sure how to fix this, but it seems to be an important problem to fix as otherwise GPJax becomes incompatible with any other libraries that rely on `plum-dispatch`, which is quite popular.

**Steps to reproduce:**

See below.

**Related code:**

It is enough to run this snippet:

```python
# Import a backend, we use jax in this example.
import jax.numpy as jnp
import jax
import gpjax as gpx

# Import the geometric_kernels backend.
import geometric_kernels
import geometric_kernels.jax
```

which leads to

```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[1], line 4
      2 import jax.numpy as jnp
      3 import jax
----> 4 import gpjax as gpx
      6 # Import the geometric_kernels backend.
      7 import geometric_kernels

File ~/anaconda3/envs/gkconda_newjax/lib/python3.10/site-packages/gpjax/__init__.py:15
      1 # Copyright 2022 The GPJax Contributors. All Rights Reserved.
      2 #
      3 # Licensed under the Apache License, Version 2.0 (the "License");
   (...)
     13 # limitations under the License.
[...]

PR #529, opened 2025-09-04
Thomaspinder/replace cola

thomaspinder:
## Checklist

## Description

This PR deprecates the use of cola.

Issue Number: N/A


thomaspinder:
> Out of interest, is this just because CoLA isn't maintained? Are you just taking the relevant code from there into gpjax.linalg?

Precisely. The awkward forking of plum in cola also causes issues e.g., #441

Issue #558, opened 2025-10-19
bug: Graph Kernel error when used within a sum kernel


syedzayyan:
# Bug Report

**GPJax version:**

0.13.0 (latest version from git)

**Current behavior:**

When the Graph Kernel is used in conjunction with another kernel using Sum function, this happens:

```
TypeError: GraphKernel.__call__() missing 1 required keyword-only argument: 'S'

File ~/Desktop/string_gp/venv/lib/python3.13/site-packages/gpjax/scan.py:161, in vscan(f, init, xs, length, reverse, unroll, log_rate, log_value)
    157     _callback(_is_last, _close_tqdm)
    159     return carry, y
...
   (...)    283         ScalarFloat: The evaluated kernel function at the supplied inputs.
    284     """
--> [285](https://file+.vscode-resource.vscode-cdn.net/Users/bty644/Desktop/string_gp/~/Desktop/string_gp/venv/lib/python3.13/site-packages/gpjax/kernels/base.py:285)     return self.operator(jnp.stack([k(x, y) for k in self.kernels]))
```

**Expected behavior:**

Normal Summation and a kernel running.

Can't do a full reproduction without preamble of graphs, hence here is a notebook:

[graph_add_kernel.ipynb](https://github.com/user-attachments/files/22989189/graph_add_kernel.ipynb)

**Other information:**

There is a To do note in the graph kernel to make the graph kernel more consistent:

```
    def __call__(  # TODO not consistent with general kernel interface
        self,
        x: Int[Array, "N 1"],
        y: Int[Array, "M 1"],
    ):
```
If the EigenComputation is discarded and instead calculated within the kernel call itself it works, like below. Perhaps there was a reason an extra computation function is required?

```
    def __call__(
        self,
        x: Int[Array, "N 1"],
        y: Int[Array, "M 1"],
    ):
        S = jnp.power(
            self.eigenvalues
            + 2
            * self.smoothness.value
            / self.lengthscale.value
            / self.lengthscale.value,
            -self.smoothness.value,
        )
        S = jnp.multiply(S, self.num_vertex / jnp.sum(S))
        # Scale the transform eigenvalues by the kernel variance
        S = jnp.multiply(S, self.variance.value)
        Kxx = (jax_gather_nd(self.eigenvectors, x) * S.squeeze()) @ jnp.transpose(
            jax_gather_nd(self.eigenvectors, y)
        )  # shape (n,n)
        return Kxx.squeeze()

```

PR #560, opened 2025-10-21
Graph kernel consistency with other kernels

syedzayyan:
## Checklist

- [x] I've formatted the new code by running `uv run poe format` before committing.

## Description

This is in reference to the issue #558 . I have created another helper function to calculate S and moved it out to utils instead. I have kept the eigen computation bits so that the tests don't fail. 

Not sure if this is the preferred way. I did think of moving the S calculations to the `_call_` function itself but it maybe required for other kernels in the future?

Issue Number: #558

### DIFF
--- a/gpjax/kernels/non_euclidean/utils.py
+++ b/gpjax/kernels/non_euclidean/utils.py
@@ -13,12 +13,19 @@
 # limitations under the License.
 # ==============================================================================
 
+from __future__ import annotations
+
+import beartype.typing as tp
+import jax.numpy as jnp
 from jaxtyping import (
     Float,
     Int,
 )
 
 from gpjax.typing import Array
+
+if tp.TYPE_CHECKING:
+    from gpjax.kernels.non_euclidean.graph import GraphKernel
 
 
 def jax_gather_nd(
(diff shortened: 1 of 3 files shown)


gpjax/kernels/non_euclidean/utils.py
```diff
@@ -41,3 +42,26 @@ def jax_gather_nd(
     """
     tuple_indices = tuple(indices[..., i] for i in range(indices.shape[-1]))
     return params[tuple_indices]
+
+
+def calculate_S(kernel):
```
thomaspinder:
```suggestion
def calculate_heat_semigroup(kernel: GraphKernel) -> Float[Array, "N N"]:
```

thomaspinder: To bypass a circular import only for the needs of types, you can do 
```
from __future__ import annotations
from import beartype.typing as tp

if tp.TYPE_CHECKING:
    from gpjax.kernels.non_euclidean.graph import GraphKernel
```

Overall this looks good. Can you please bump the version in `gpjax/__init__.py`?

syedzayyan: You're a genius! Worked like a charm. Function is back in utils and bumped __init__

thomaspinder (approved):

Issue #566, opened 2025-10-30
bug: Compute Engine Diagonal is returning the wrong type


mathDR:
In `GPJax/gpjax/kernels.base.py`, the return type of the `diagonal` method in `AbstractKernel` should change from
```
def diagonal(self, x: Num[Array, "N D"]) -> Float[Array, " N"]:
```
to
```
def diagonal(self, x: Num[Array, "N D"]) -> LinearOperator:
```
to both match the return type of `gram()` and the expected return type in `AbstractKernelComputation`.

PR #567, opened 2025-11-01
Ability to return just diag on `predict` call

### LINKED ISSUE #389: dev: Improve GP Prediction Efficiency
Currently when calling `predict` on a Gaussian process GPJax computes the entire covariance matrix. However, when calling with many test points this can be inefficient for downstream tasks which only require the mean and variance at the test points (and hence only require the diagonal elements of the covariance matrix). It would be good to come up with a solution which avoids unnecessary computation of elements in the covariance matrix. For instance, perhaps with some sort of lazy evaluation elements in the covariance matrix could only be computed when necessary. Alternatively, there could be an efficient way of doing this using [Cola](https://github.com/wilson-labs/cola).

mathDR:
## Checklist

- [X] I've formatted the new code by running `uv run poe format` before committing.

## Description
This PR does one thing:  allows for the user to request to _only_ return the diagonal elements of the covariance matrix of the test points.

This is done through the addition of a `typing.Literal` named `return_cov_type` that is passed into the respective `predict` functions for `Prior`, `ConjugatePosterior` and `NonConjugatePrior`.  When the literal is `"dense"` then the full covariance is computed and returned.  If the literal is `"diagonal"` then _only_ the diagonal elements are computed and returned.

The implementation of this is through a `jax.lax.cond` which is the jax equivalent of an `if-else` capability.  One drawback of this architecture is that both branches of the `cond` must return the same type of `PyTree`.  Therefore in returning the covariance on the `diagonal` branch, we must cast the diagonal array to a `Dense` type `LinearOperator`.

In implementing this PR, some issues with the `Diagonal` functionality in `kernels.computations` were discovered and fixed.

This PR closes #389 and #566

### DIFF
--- a/gpjax/gps.py
+++ b/gpjax/gps.py
@@ -150,8 +150,8 @@ class Prior(AbstractPrior[M, K]):
     r"""A Gaussian process prior object.
 
     The GP is parameterised by a
-    [mean](https://docs.jaxgaussianprocesses.com/api/mean_functions/)
-    and [kernel](https://docs.jaxgaussianprocesses.com/api/kernels/base/)
+    [mean](https://docs.thomaspinder.com/api/mean_functions/)
+    and [kernel](https://docs.thomaspinder.com/api/kernels/base/)
     function.
 
     A Gaussian process prior parameterised by a mean function $m(\cdot)$ and a kernel
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,1 @@
+github: [thomaspinder]
--- a/.github/workflows/pr_greeting.yml
+++ b/.github/workflows/pr_greeting.yml
@@ -35,7 +35,7 @@ jobs:
 
             We strive to be a welcoming and open project. Please follow our
             [Code of
-            Conduct](https://github.com/JaxGaussianProcesses/GPJax/blob/main/.github/CODE_OF_CONDUCT.md).
+            Conduct](https://github.com/thomaspinder/GPJax/blob/main/.github/CODE_OF_CONDUCT.md).
 
           pr_message: |
             Thank you for opening your first PR into GPJax!
@@ -59,4 +59,4 @@ jobs:
 
             We strive to be a welcoming and open project. Please follow our
             [Code of
-            Conduct](https://github.com/JaxGaussianProcesses/GPJax/blob/main/.github/CODE_OF_CONDUCT.md).
+            Conduct](https://github.com/thomaspinder/GPJax/blob/main/.github/CODE_OF_CONDUCT.md).
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,22 +3,22 @@ name: "Release Automation"
 on:
   push:
     tags:
-      - 'v*'
+      - "v*"
   workflow_dispatch:
     inputs:
       version:
-        description: 'Version to release (e.g., v0.6.0)'
+        description: "Version to release (e.g., v0.6.0)"
         required: true
         type: string
       draft:
-        description: 'Create draft release'
+        description: "Create draft release"
         required: false
         default: false
         type: boolean
 
 permissions:
   contents: write
-  id-token: write  # For PyPI trusted publishing
+  id-token: write # For PyPI trusted publishing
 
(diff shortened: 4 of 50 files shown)


mathDR:
Oh I suppose I should add a test checking that things error out appropriately when calling with an incorrect literal...

gpjax/gps.py
```diff
@@ -13,10 +13,16 @@
 # limitations under the License.
 # ==============================================================================
 # from __future__ import annotations
+
 from abc import abstractmethod
+from typing import (
+    Literal,
```
thomaspinder: Types should be imported from beartype.

gpjax/gps.py
```diff
@@ -70,14 +80,19 @@ def __init__(
         r"""Construct a Gaussian process prior.
 
         Args:
-            kernel: kernel object inheriting from AbstractKernel.
-            mean_function: mean function object inheriting from AbstractMeanFunction.
+        kernel: kernel object inheriting from AbstractKernel.
+        mean_function: mean function object inheriting from AbstractMeanFunction.
```
thomaspinder: Did you run a linter/formatter on the code? It's odd to see changes like this coming in. The linting workflow will not error as it is not possible to lint docstrings, but I would like the format to be consistent across the codebase, so can you please undo this formatting?

gpjax/gps.py
mathDR: beartype doesn't support Literal.  In their [documentation](https://beartype.readthedocs.io/en/latest/tldr/) they state explicitly to use `typing.Literal`

gpjax/gps.py
```diff
@@ -239,19 +275,44 @@ def predict(self, test_inputs: Num[Array, "N D"]) -> GaussianDistribution:
...
+        def _ret_full_cov(
+            t: Num[Array, "N D"],
+        ) -> Tuple[Float[Array, " N"], LinearOperator]:
+            mean_at_test = self.mean_function(t)
+            Kxx = self.kernel.gram(t)
+            Kxx_dense = add_jitter(Kxx.to_dense(), self.jitter)
+            Kxx = psd(Dense(Kxx_dense))
+            return jnp.atleast_1d(mean_at_test.squeeze()), Kxx
+
+        def _ret_diag_cov(
```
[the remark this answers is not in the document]
mathDR: Good call.  I refactored the one in `Prior` since it was one line to compute the mean, but renamed the ones in `Posterior` since the mean calculation was a little more involved.

Since the mean computation is the same for both return branches, I am happy to pull that part out and refactor the branch functions to only return the respective covariance.  Just let me know your preference.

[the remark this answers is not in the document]
thomaspinder: Yes - the refactoring you propose here feels cleanest. Let's only return the covariance.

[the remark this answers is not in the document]
mathDR: So the mean function actually depends upon values from the covariance.  Since the full covariance branch has a `Dense` representation and the `diagonal` branch has a `Diagonal` one, I propose to have the respective functions compute their respective mean and covariance within the branch.

gpjax/gps.py
thomaspinder: TIL... Thanks!

mathDR:
Generically I am refactoring this and adding more tests.  i see that in the [scipy](https://github.com/scikit-learn/scikit-learn/blob/c60dae20604f8b9e585fc18a8fa0e0fb50712179/sklearn/gaussian_process/_gpr.py#L443-L496) version, they pre-compute some `Dense` matrices to enable mean computations outside of the covariance computations.

In our case, can take advantage of computing the Cholesky of `Kxx` and the respective solves we need for those, so I hesitate to pull that piece out of each branch.

Also, we must be calling `vmap` on this predict function, because my tests are showing that _both_ branches of the `jax.lax.cond` are being computed (which defeats the whole purpose! :rofl: )  So I will try to run that down...

mathDR:
> I just opened a PR to your branch that fixes the graph kernel issue. If we first merge that, then the tests here should pass and we can merge into main.

~weird that my `main` branch has all of the `add_diag_cov` branch changes.  Let me rebase then I can review your PR.~

disregard this is how forks work.  🤦

PR #575, opened 2025-11-16
Multioutput proposal

mathDR:
# MultiOutput GP Proposal.

In the spirit of [MultiOutput Processes like GPFlow](https://gpflow.github.io/GPflow/2.4.0/notebooks/advanced/multioutput.html), we will take the same problem statement:

## Problem Statement
We will consider a regression problem for functions $f: \mathbb{R}^D \rightarrow \mathbb{R}^Q$. We assume that the dataset is of the form $(X_1, f_1), \dots, (X_Q, f_Q)$, that is, we may observe _different_ inputs for each output dimension.

Here we assume a model of the form:
$$f(x) = W g(x), $$
where $g(x) \in \mathbb{R}^L$, $f(x) \in \mathbb{R}^Q$ and $W \in \mathbb{R}^{Q \times L}$. We assume that the outputs of $g$ are uncorrelated, and that by *mixing* them with $W$ they become correlated. 

Note, we have two options for $g$:
1. The output dimensions of $g$ share the _same_same kernel.
2. Each output of $g$ has a _separate_ kernel.

In a following PR we can discuss variational GPs, wherein we need to further suboption for the inducing inputs of $g$:
1. The instances of $g$ share the same inducing inputs.
2. Each output of $g$ has its own set of inducing inputs.

The notation is as follows:
 - $X_i \in \mathbb{R}^{N \times D}, i=1,ldots,Q$ denotes the input
 - $Y \in \mathbb{R}^{N \times Q}$ denotes the output
 - $k_{1..L}$, $L$ are kernels on $\mathbb{R}^{N \times D}$
 - $g_{1..L}$, $L$ are independent GPs  with $g_l \sim GP(0,k_l)$
 - $f_{1..Q}$, $P$ are correlated  GPs  with $\bf{f} = \bf{W} \bf{g}.$

## Phase 1:
We write a multioutput kernel that initially just takes in a single kernel.  Mimicing GPFlow again we could have something like:
```
"""MultiOutput Kernels."""

from abc import abstractmethod

import jax.numpy as jnp
from jaxtyping import Num

from gpjax.kernels import AbstractKernel
from gpjax.typing import Array


# TODO describe various output shapes
class MultioutputKernel(AbstractKernel):
    """
    Multi Output Kernel class.

    This kernel can represent correlation between outputs of different datapoints.

    The `full_output_cov` argument holds whether the kernel should calculate
    the covariance between the outputs. In case there is no correlation but
    `full_output_cov` is set to True the covariance matrix will be filled with zeros
    until the appropriate size is reached.
    """

    @property
    @abstractmethod
    def num_latent_gps(self) -> int:
        """The number of latent GPs in the multioutput kernel"""
        raise NotImplementedError

    @property
    @abstractmethod
    def latent_kernels(self) -> tuple[AbstractKernel, ...]:
        """The underlying kernels in the multioutput kernel"""
        raise NotImplementedError

    @abstractmethod
    def K(
        self,
        X: Num[Array, "N D"],
        X2: Num[Array, "M D"] | None = None,
        full_output_cov: bool = True,
    ) -> Array:
        """
        Returns the correlation of f(X) and f(X2), where f(.) can be multi-dimensional.

        :param X: data matrix
        :param X2: data matrix
        :param full_output_cov: calculate correlation between outputs.
        :return: cov[f(X), f(X2)]
        """
        raise NotImplementedError

    @abstractmethod
    def K_diag(self, X: Num[Array, "N D"], full_output_cov: bool = True) -> Array:
        """
        Returns the correlation of f(X) and f(X), where f(.) can be multi-dimensional.

        :param X: data matrix
        :param full_output_cov: calculate correlation between outputs.
        :return: var[f(X)]
        """
        raise NotImplementedError

    def __call__(
        self,
        X: Num[Array, "N D"],
        X2: Num[Array, "M D"] | None = None,
        *,
        full_cov: bool = False,
        full_output_cov: bool = True,
    ) -> Array:
        if not full_cov and X2 is not None:
            raise ValueError(
                "Ambiguous inputs: passing in `X2` is not compatible with `full_cov=False`."
            )
        if not full_cov:
            return self.K_diag(X, full_output_cov=full_output_cov)
        return self.K(X, X2, full_output_cov=full_output_cov)


class SharedIndependent(MultioutputKernel):
    """
    - Shared: we use the same kernel for each latent GP
    - Independent: Latents are uncorrelated a priori.
    """

    def __init__(self, kernel: AbstractKernel, output_dim: int) -> None:
        super().__init__()
        self.kernel = kernel
        self.output_dim = output_dim

    @property
    def num_latent_gps(self) -> int:
        # In this case number of latent GPs (L) == output_dim (P)
        return self.output_dim

    @property
    def latent_kernels(self) -> tuple[AbstractKernel, ...]:
        """The underlying kernels in the multioutput kernel"""
        return (self.kernel,)

    def K(
        self,
        X: Num[Array, "N D"],
        X2: Num[Array, "M D"] | None = None,
        full_output_cov: bool = True,
    ) -> Array:
        if X2 is None:
            K = self.kernel.gram(X)
        else:
            K = self.kernel.cross_covariance(X, X2)

        if full_output_cov:
            return jnp.tile(
                jnp.tile(K, (self.num_latent_gps, 1)), (1, self.num_latent_gps)
            )
        else:
            return jnp.kron(jnp.eye(self.num_latent_gps, dtype=int), K)

    def K_diag(self, X: Num[Array, "N D"], full_output_cov: bool = True) -> Array:
        K = self.kernel.diagonal(X)
        Ks = jnp.tile(K, (self.num_latent_gps, 1))
        return jnp.diag(Ks) if full_output_cov else Ks
```
we could then add another `Independent` Kernel where we pass in a `Sequence` of kernels and apply each to a `num_latent_gps` dimension.

## Phase 2
We add a `MultiLatentPrior` class (probably just start with Conjugate) where instead of returning a `GaussianDistribution` class, we develop a `MatrixNormalDistribution` class that inherits from numpyro's [MatrixNormal](https://num.pyro.ai/en/stable/distributions.html#matrixnormal) that will allow us to sample.

I am still fleshing this out, but overall it seems as if we need to add a Multikernel and a multi prior of some types.

Thoughts?

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -61,7 +61,7 @@ jobs:
           uv run mkdocs build
 
       - name: Deploy Page 🚀
-        uses: JamesIves/github-pages-deploy-action@v4.7.3
+        uses: JamesIves/github-pages-deploy-action@v4.7.4
         with:
           branch: gh-pages
           folder: site
--- a/examples/regression.py
+++ b/examples/regression.py
@@ -121,10 +121,15 @@ prior = gpx.gps.Prior(mean_function=meanf, kernel=kernel)
 # we have just defined can be represented by a
 # [TensorFlow Probability](https://www.tensorflow.org/probability/api_docs/python/tfp/substrates/jax)
 # multivariate Gaussian distribution. Such functionality enables trivial sampling, and
-# the evaluation of the GP's mean and covariance .
+# the evaluation of the GP's mean and covariance.
+#
+# Since we want to sample from the full posterior, we need to calculate the full covariance matrix.
+# We can enforce this by including the `return_covariance_type = "dense"` attribute when predicting.
+# Note this is what will be defaulted if left blank.
 
 # %%
-prior_dist = prior.predict(xtest)
+# %% [markdown]
+prior_dist = prior.predict(xtest, return_covariance_type="dense")
 
 prior_mean = prior_dist.mean
 prior_std = prior_dist.variance
@@ -212,9 +217,13 @@ print(-gpx.objectives.conjugate_mll(opt_posterior, D))
 # this, we use our defined `posterior` and `likelihood` at our test inputs to obtain
 # the predictive distribution as a `Distrax` multivariate Gaussian upon which `mean`
 # and `stddev` can be used to extract the predictive mean and standard deviatation.
+# 
+# We are only concerned here about the variance between the test points and themselves, so
+# we can just copute the diagonal version of the covariance.  We enforce this by using
+# `return_covariance_type = "diagonal"` in the `predict` call.
 
 # %%
-latent_dist = opt_posterior.predict(xtest, train_data=D)
+latent_dist = opt_posterior.predict(xtest, train_data=D, return_covariance_type="diagonal")
 predictive_dist = opt_posterior.likelihood(latent_dist)
 
 predictive_mean = predictive_dist.mean
--- a/gpjax/__init__.py
+++ b/gpjax/__init__.py
@@ -40,7 +40,7 @@ __license__ = "MIT"
 __description__ = "Gaussian processes in JAX and Flax"
 __url__ = "https://github.com/thomaspinder/GPJax"
 __contributors__ = "https://github.com/thomaspinder/GPJax/graphs/contributors"
-__version__ = "0.13.2"
+__version__ = "0.13.3"
 
 __all__ = [
     "gps",
--- a/gpjax/mean_functions.py
+++ b/gpjax/mean_functions.py
@@ -147,9 +147,9 @@ class Constant(AbstractMeanFunction):
             Float[Array, "1"]: The evaluated mean function.
         """
         if isinstance(self.constant, Parameter):
-            return jnp.ones((x.shape[0], 1)) * self.constant.value
+            return jnp.ones((x.shape[0], 1), dtype=x.dtype) * self.constant.value
         else:
-            return jnp.ones((x.shape[0], 1)) * self.constant
+            return jnp.ones((x.shape[0], 1), dtype=x.dtype) * self.constant
 
 
 class Zero(Constant):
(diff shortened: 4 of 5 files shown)


mathDR:
In the general case with $L$ latent kernels, we need to compute:
$$\bf{vec}(y) \sim \mathbb{N}(\bf{vec}(0), \Sigma + \sigma^2 \bf{I}_{NQ}), \Sigma = \sum_l K_l \otimes w_l w_l^T$$

mathDR:
Also, not for this PR, but when we allow for variational inference, we will have a new `get_posterior()` method that (probably) matches something like what [GPFlow](https://github.com/GPflow/GPflow/blob/6f48b0d32d8e87d043035449936d72549632342b/gpflow/posteriors.py#L1042-L1089) has in this area:

So we would have a `SharedIndendent`, a `SeperateIndependent`, and a `LinearCoregionalization` multioutput set of kernels.  And we would have a matching prior for each.

For the variational model(s) we would need to specify for the latter two kernels if we want the inducing points "shared" among all of the kernels, or each kernel having its own set of inducing points.

But as I said, that is for a later PR.

mathDR:
For this PR:  we would implement:
1. MultiOutputKernels
a. SharedIndependent
b. SeperateIndependent
c. LinearCoregionalization
2. Prior/Posteriors for the above.  We assume only conjugate priors for this PR (a later PR would add the nonconjugate case). There would be a one-to-one match between them, so:
a. SharedIndependentPrior; SharedIndependentConjugatePosterior
b. SeperateIndependentPrior; SeperateIndependentConjugatePosterior
c. LinearCoregionalizationPrior; LinearCoregionalizationPosterior


Thoughts?

thomaspinder:
OK. I'm good with leaving the mixing component in the prior, this is fine. I also align with your proposal for this PR.

thomaspinder:
Closing this in favour of #604 / #605 / #606

PR #634, opened 2026-04-30
Kronecker mv fix

lugin100:
## Checklist

- [✓] I've formatted the new code by running `uv run poe format` before committing.
- [✓] I've added tests for new code.
- [] I've added docstrings for the new code.

## Description
While using GPJax I encountered a bug in the Kronecker linear operator in /linalg/custom_operators.py: 
When the passed linear operators are not square, the matrix vector product will fail with a TypeError. This is because the function tries to reshape the input vector to the shape of the output. This PR adresses this by reshaping the input to the expected input shapes instead.

I have added a test for this case and have run poe format.


thomaspinder (approved):

thomaspinder:
Failing check is a Github CI/CD, and issue (Issue #635 ) will resolve. Thanks @lugin100 !

PR #769, opened 2026-08-11
Simplify bijection logic by solely using val() to tranform parameters

Thomas-Christie:
## Checklist

- [X] I've formatted the new code by running `uv run poe format` before committing.
- [N/a] I've added tests for new code.

## Description

This change simplifies the handling of bijections. Parameter conversion now follows a simple rule - if you want to perform arithmetic on a parameter (normally in the constrained space), simply call `val()` on the parameter. There is no need to call `paramax.unwrap()` on a model. This commit enforces this rule by:
1. Making [_val](https://github.com/thomaspinder/GPJax/blob/main/gpjax/parameters.py#L57) public.
2. Removing the [paramax.unwrap](https://github.com/thomaspinder/GPJax/blob/main/gpjax/fit.py#L117) of models in `fit` methods - parameter constraints are already handled correctly *within* objectives, where parameters are constrained when accessed, via `val`.
3. Removing the remaining model-level unwraps in `gpjax/state_space/{objectives,prediction}.py` and `gpjax/natural_gradients.py`.
4. Altered docs/tests to respect this philosophy.

All tests still pass. The main danger of things breaking would be from the removal of `paramax.unwrap()` in the `fit` methods, but in such cases, this would manifest in errors when the objective was called as arithmetic operations would be attempted on `AbstractUnwrappable`. No such instances of these errors occur.

NOTE: `with_log_prior` has been removed for now - the way this was written had the implicit assumption that the model passed to it was in an *unwrapped* state. I will be rewriting this for the new philosophy in the next PR.

### DIFF
--- a/docs/sharp_bits.md
+++ b/docs/sharp_bits.md
@@ -98,16 +98,100 @@ parameters during optimisation. Each constrained parameter — [`PositiveReal`](
 Equinox-compatible pytree node whose `unwrap()` method applies the constraining
 bijection (e.g. softplus for positivity, sigmoid for bounded parameters).
 
-During optimisation, [`fit`](#gpjax.fit.fit) calls
-[`paramax.unwrap`](inv:paramax#paramax.wrappers.unwrap) inside the loss function.
-This recursively resolves every `AbstractUnwrappable` leaf in the model tree, mapping
-internal unconstrained values to their constrained counterparts. Gradients are computed
-in the unconstrained space, and updates are applied directly to the unconstrained
-arrays — no explicit forward/inverse transform step is needed.
+Gradients are computed in the unconstrained space, and updates are applied directly to
+the internal unconstrained arrays — no explicit forward/inverse transform step is
+needed, and [`fit`](#gpjax.fit.fit) never has to convert your model into a different
+form to optimise it.
+
+### When do I need to call `val()`?
+
+Models are **always** in their wrapped form. A model you build is wrapped, and a model
+returned by `fit` is still wrapped — nothing in GPJax converts it to a plain-array
+model, and you never need to do so yourself. The single rule is:
+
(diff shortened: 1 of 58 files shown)


docs/sharp_bits.md
```diff
@@ -98,16 +98,100 @@ parameters during optimisation. Each constrained parameter — [`PositiveReal`](
...
-This recursively resolves every `AbstractUnwrappable` leaf in the model tree, mapping
-internal unconstrained values to their constrained counterparts. Gradients are computed
-in the unconstrained space, and updates are applied directly to the unconstrained
-arrays — no explicit forward/inverse transform step is needed.
+Gradients are computed in the unconstrained space, and updates are applied directly to
+the internal unconstrained arrays — no explicit forward/inverse transform step is
+needed, and [`fit`](#gpjax.fit.fit) never has to convert your model into a different
+form to optimise it.
+
+### When do I need to call `val()`?
```
thomaspinder, line 106: This is a helpful addition. Thanks

gpjax/objectives.py
```diff
@@ -32,92 +32,6 @@
 LogPriorFn = tpe.Callable[[eqx.Module], ScalarFloat]
 
 
-def with_log_prior(objective: Objective, log_prior: LogPriorFn) -> Objective:
```
thomaspinder, line 35: Did you intentionally remove?

thomaspinder (approved):

gpjax/objectives.py
Thomas-Christie: Yes - the way some of the unit tests had been written for it had the implicit assumption that the model passed to `_regularised_objective` was in an unwrapped state, and so were failing with my refactor, due to `fit` no longer unwrapping models before training. Given that I'll be rewriting the `with_log_prior` logic in the next PR, and this isn't in `main` yet, I thought it made sense to remove rather than get the unit tests running for something I am about to change.

